Issue #76, opened 2020-07-02
Add crypto "test parameters" to speed up unit tests


danwallach:
**The problem**: Big unit tests suites run slowly, bottlenecking on the cryptographic computations.

**The current workaround**: Reducing the number of tests or the size of the tests that run. Particularly for the Hypothesis tests, this reduces their ability to find bugs.

**The better solution**: Have smaller values for `P`, `Q`, and `G`, such that everything fits in 32 bits or even 16 bits. 

**The challenge**: How do we introduce "test parameters" without them ever, ever finding their way into production code? What's the minimal set of changes to add a "test mode"?

Two broad options:
1) "Monkey patching" or otherwise having test code, from outside of the `electionguard` package, reach into the package and make changes. Editing the values of `P`, `Q`, and `G` is easy. Similarly, it looks like all you have to do is change `Q_MINUS_ONE` and you'll fix `hash_elems`.
2) Having a "test mode switch" wired directly into the `electionguard` package. It sets a globally visible boolean flag and overwrites the values of `P`, `Q`, `G`, etc. It would also nuke the `dlog` cache state.

I'm slightly in favor of the "test mode", since it can generate suitably loud warnings if it's ever used in a real election, or could even disable specific features, like writing persistent data to a file. Another way to ensure mistakes aren't made is to just write out the parameters as part of any election file, so if we see any election with "test parameters", that will be an instant read flag. Like, trying to load up some persistent data, the first thing you do is check the parameters, and if they're not equal to what you expect, you raise an error.

**How do you pick the right parameters?** Says our resident cryptographer: "All we need is any small prime q that is larger than the number of voters and a larger prime p such that p-1 is a multiple of q.  We don’t have to go to all of the trouble I did to select primes to optimize other characteristics and to do so in a publicly-justifiable way."

I already had a Sieve of Eratosthenes program hanging around, so I searched for the smallest `Q` greater than 32767 (2^15 - 1) and came up with:
```
P = 65543
Q = 32771
```

danwallach:
I'm playing with this from external code, and one issue is that P and Q are final, so assignments to them are "not allowed" (to the extent that anything in Python is not allowed).

danwallach:
Update: as an experiment, I just put these small constants into `group.py` and tried running the unit test suite. I tried G=4 (what you get by default) and G=2 (Josh had suggested). They yielded different tests failing.

Further examination of this led me to discover one bug in test_elgamal.py:
```
# starts at line 87
[...]

PR #77, opened 2020-07-02
🗳 Election Artifacts

### LINKED ISSUE #14: Output Data Representation
Determine the proper format for representing output data from the system.

This was brought up as part of this: https://github.com/microsoft/electionguard-python/pull/1#discussion_r412531622

One consideration is the representation of `ElementModP` and `ElementModQ` which will require some custom parsing to properly represent these values

keithrfung:
### Issue
Fixes #14 

### Description
This outputs the data from the election at the end.  This data will later be used as artifacts to verify the election. 
Screen Shot 2020-07-02 at 3 34 22 PM

### Testing
This can be best seen from the integration test. 

### Checklist


src/electionguard/election.py
```diff
@@ -773,6 +773,19 @@ def _generate_contests_with_placeholders(
         return contests
 
 
+@dataclass(frozen=True)
+class ElectionConstants(Serializable):
+    """
+    The constants for mathematical functions during the election. 
+    """
```
danwallach, line 780: I really like this. An interesting thing to ponder is rummaging around the rest of the codebase to make sure everybody uses these parameters rather than the global constants. That would be really beneficial to support "test mode" parameters. (https://github.com/microsoft/electionguard-python/issues/76)

src/electionguard/publish.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+from .ballot_box import CiphertextBallot
+from .decryption_mediator import PlaintextTally
+from .election import CiphertextElectionContext, ElectionConstants, ElectionDescription
+from .key_ceremony import CoefficientValidationSet
+from .tally import CiphertextTally
+from .group import ElementModP, ElementModQ
+
+RECORD_DIR = "output" + path.sep
```
danwallach, line 14: I don't know if this is a great default value, although hopefully the caller of `publish` will specify something better.

src/electionguard/publish.py
```diff
@@ -0,0 +1,78 @@
...
+
+COEFFICIENT_PREFIX = "coefficient_validation_set_"
+BALLOT_PREFIX = "ballot_"
+
+
+def publish(
+    description: ElectionDescription,
+    context: CiphertextElectionContext,
+    constants: ElectionConstants,
+    coefficient_validation_sets: List[CoefficientValidationSet],
```
danwallach, line 33: In some cases, there will be no coefficients hanging around because it's just a single trustee. Should this list be empty in those cases? `Optional`?

src/electionguard/publish.py
```diff
@@ -0,0 +1,78 @@
...
+
+    make_directory(COEFFICIENTS_DIR)
+    for coefficient_validation_set in coefficient_validation_sets:
+        set_name = COEFFICIENT_PREFIX + coefficient_validation_set.owner_id
+        coefficient_validation_set.to_json_file(set_name, COEFFICIENTS_DIR)
+
+    make_directory(BALLOTS_DIR)
+    for ballot in ciphertext_ballots:
+        ballot_name = BALLOT_PREFIX + ballot.object_id
+        ballot.to_json_file(ballot_name, BALLOTS_DIR)
```
danwallach, line 56: If you've got a million ballots, which could totally happen, having a million files in a single directory is going to be a performance problem every time you try to just list the files. There needs to be some sort of subdirectory structure here.

src/electionguard/publish.py
```diff
@@ -0,0 +1,78 @@
...
+    description: ElectionDescription,
+    context: CiphertextElectionContext,
+    constants: ElectionConstants,
+    coefficient_validation_sets: List[CoefficientValidationSet],
+    ciphertext_ballots: List[CiphertextBallot],
+    ciphertext_tally: CiphertextTally,
+    plaintext_tally: PlaintextTally,
+    record_directory: str = RECORD_DIR,
+) -> None:
+    """Publishes the election record as json"""
```
danwallach, line 39: These JSON files will be quite amenable to compression (gzip or otherwise), and there are a number of Python libraries that can help with this. Given the potentially quite large amount of disk space being used here, you should consider compression.

danwallach:
One more feature that would be really nice to have here: a manifest of hash values for each file, which can then be signed by an election official using conventional cryptography. Probably the right place to start looking for a suitable design is the Java JAR manifest specification (which, believe it or not, I was "in the room where it happened" when we came up with this thing).

- https://docs.oracle.com/javase/7/docs/technotes/guides/jar/jar.html

We don't have to clone that, but it gives you a sense of how something like this might work. This would allow for incremental verification when the files are being read back in, which could be done incrementally as necessary.

src/electionguard/publish.py
keithrfung: This is a piece we're looking at in a future card. The point is to provide a tool that individuals can use to output and a framework for this. We're more focusing on the file itself not the storage process in this PR.

Especially because developers can choose to change how these files are output.

src/electionguard/publish.py
keithrfung: This looks at extendability that we want to open to the users of the libraries and not confine them to. Although we can we want to minimize dependencies.

src/electionguard/publish.py
keithrfung: Fair point. Waiting to see what the team puts on this. Suggestions welcome.

src/electionguard/publish.py
keithrfung: We can make the parameter optional by defaulting to empty list.

src/electionguard/publish.py
keithrfung: Other options considered were `artifact(s)` or `record(s)`

src/electionguard/publish.py
danwallach: The standard CPython distribution includes gzip support (https://docs.python.org/3/library/archiving.html). All of that bottoms out at `zlib`, which is written in C and included along with CPython. As such, we're not really adding any new dependencies. What we are doing, however, is creating more attack surface.

Ideally, we'd have our own binary format (msgpack, protobuf, etc.), but I imagine that's not on the short-term roadmap.

src/electionguard/publish.py
AddressXception: yeah i think output could generally be considered a word reserved for build state in some cases.  my vote is for `results`

src/electionguard/publish.py
AddressXception: let's put this in a new issue so we can more properly consider outside the scope of this PR.  binary serialization is on the road map and i think we need to look at all of that together

src/electionguard/serializable.py
```diff
@@ -1,35 +1,58 @@
 from dataclasses import dataclass
-from typing import cast, TypeVar, Generic
-
 from jsons import KEY_TRANSFORMER_CAMELCASE, KEY_TRANSFORMER_SNAKECASE, dumps, loads
+from typing import cast, TypeVar, Generic
 
 T = TypeVar("T")
 
+JSON_FILE_EXTENSION: str = ".json"
+WRITE: str = "w"
```
AddressXception, line 8: is this consistent across platforms?

keithrfung: This is the python setting so yes.

PR #99, opened 2020-07-10
Secret Sharing in RSA

### LINKED ISSUE #35: Encrypt `extended_data`
The `PlaintextBallotSelection` includes an `extended_data` field that should be encrypted and decrypted.

Question: The encryption method should be able to support string lengths up to 256? or 256 bits?

In order to make the ballots homogenous, extendedData values are wrapped in an internal data structure and padded to max length. The first 4 characters are reserved for the length, and then the string follows. 
 if no data is passed in the extended data field, then random data is generated to fill the field and it is keyed specifically to be identified as random data.

For instance, something like this:

1. extendedData in: `some write in candidate name`, extended data out: `028:some write in candidate name_gj8749hg8g55h98...` where 028 is the character count of the real data
2. extendedData In: `(an empty string)`, extendedData out: `000:_jh7453h7egnu9...`

as an alternative we could use ore reuse some binary serialization mechanism and then encrypt the serialized value?

suggestions welcome

### LINKED ISSUE #47: Implement Master Ballot Nonce Encryption
there are several cases where we need to use a free-form encryption in the application.

1. when sharing keys between guardians
2. when storing or sharing nonce values
3. when encrypting/decrypting value in the `extended_data` field on a ballot

we will implement an "ordinary elgamal" encryption that will give us up to `3840` bits fo fidelity.

we will choose a prime p for ordinary ElGamal such that p mod 4 = 3, option 3 becomes easier in this case because it ensures that (-1) is a non-square mod p.

say we have a message `m`:

1.  append SHA-256(m) to get M=(m|SHA-256(m))
2.  Check if M is a square mod p by computing z = M^((p-1)/2) mod p
3. If z=1, compute an ElGamal encryption of M.  If z=-1, compute an ElGamal encryption of (-M) instead.

After decrypting, check if the final 256 bits is the SHA-256 hash of the first 3840 bits.  If it is, take the first 3840 bits as m.  If not, negate the result and then take the first 3840 bits as m (the remainder should match the SHA-256 hash of m).

## Design Considerations

sensitive values should be created and immediately encrypted in the same function where they are generated. They should only be long-lived in memory in the encrypted form.

open question: Should A guardian's election key pair should also be encrypted with the free-form encryption?

We must designate a fixed-length of the data that is going to be encrypt-able by the free-form encryption.  Given that secret keys are in the space [1,Q) it is likely we should choose 256 bits

Guardian keys:
1. election public/private key pair (for exponential elgamal)
2. election free-form public/private key pair (for ordinary elgamal, encrypting nonces, extended data, etc.)
3. secret-sharing public/private key pair (for any encryption scheme, our implementation is ordinary elgamal)

Election Keys:
1. election joint public key (exponential)
2. election joint free-form public key (ordinary)

### Generating the joint free-form key

The joint free form key is generated using the exact same secret sharing process as is used for the primary election key. 

### Sharing keys between guardians

When sharing keys between guardians, the free-form encryption is used to encrypt the election private key shares and the election free-form private key shares.  The function used to do this is pluggable to support an external system choosing to use a different form of encryption, such as RSA.  A stub for this already exists.  Our default implementation will re-use the same ordinary elgamal encryption mechanism we are using for the joint free-form key.  This means we will be using an ordinary elgamal key pair in the context of a guardian secret-sharing to encrypt another ordinary elgamal key pair used in the context of encrypting free-form values in the election

### Storing or sharing nonce values

Once the guardians have shared and verified the keys, a secondary election free-form joint public key is associated with the `CiphertextElectionContext` and used to encrypt the `random master nonce` values that are generated for each ballot.  The encrypted representation of the random master nonce replaces the current unencrypted nonce on the CiphertextBallot.

To support key flexibility, we must also support the use case that a specific consumer may not be using the guardian key ceremony artifacts and therefore we must allow a single public/private key pair to serve as the election joint free-form key.  This should work in a similar format to encrypting/decrypting ballots by using a single key that bypasses the key ceremony entirely.

This supports several use cases, specifically:
1. using the key ceremony for both encrypting ballots and nonce values, but separating the contexts with the 2 separate keys described above
2. using the key ceremony for encrypting ballot selections, but disallowing the guardians from decrypting or knowing the nonce values used for the ballots
3. similarly, using a single key pair for the ballots, but using the key ceremony for the nonces
4. using single values for the ballots and using single values for the nonces

### Encrypting/decrypting the value in ExtendedData

On a ballot, the extended data field is also encrypted with the election joint free-form public key.  In order to make the ballots homogenous, extendedData values are padded to max length.  if no data is passed in the extended data field, then random data is generated to fill the field.  There is more info in #35 

In the default workflow for an end-to-end election, the key ceremony guardians are used to generate both the joint public key and the free-form public key; and the free-form key is used to encrypt both nonces and extended data.  

In our implementation, if the key ceremony guardians are used for all keys, we only support reusing the free-form key for both extended data and nonces.  The option also exists to use a different free-form key for either the nonce values or the extended data fields by providing a single key-pair for either case. It is up to the consumer to implement this if desired.

keithrfung:
### Issue

Fixes #99  

### Description
Add RSA encryption as default method to messages between guardians for sharing of the secrets. PR originally encompassed other cards hence the branch name is a bit off. There was also a bug discovered with the test_decryption_mediator.

### Testing
* test_rsa.py

### Checklist

💚Thank you!

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,18 +7,19 @@ verify_ssl = true
 black = "*"
 coverage = "*"
 docutils = "*"
+hypothesis = "==5.15.1"
 mkdocs = "*"
 mypy = "*"
 pydeps = "*"
 pydocstyle = "*"
 pylint = "*"
 pytest = "*"
+typish = '*'
 
 [packages]
-hypothesis = "==5.15.1"
 numpy = '==1.18.2'
 jsons = '==1.1.2'
-typish = '*'
+cryptography = "*"
 
 [requires]
 python_version = "3.8"
--- a/src/electionguard/auxiliary.py
+++ b/src/electionguard/auxiliary.py
@@ -1,0 +1,43 @@
+from typing import Callable, Optional, NamedTuple
+
+from .types import GUARDIAN_ID
+
+MESSAGE = str
+PUBLIC_KEY = str
+SECRET_KEY = str
+ENCRYPTED_MESSAGE = str
+
+
+class AuxiliaryKeyPair(NamedTuple):
+    """A tuple of a secret key and public key."""
+
+    secret_key: SECRET_KEY
+    """The secret or private key"""
+    public_key: PUBLIC_KEY
+
+
+class AuxiliaryPublicKey(NamedTuple):
+    """A tuple of auxiliary public key and owner information"""
+
+    owner_id: GUARDIAN_ID
+    """
+    The unique identifier of the guardian
+    """
+
+    sequence_order: int
+    """
+    The sequence order of the auxiliary public key (usually the guardian's sequence order)
+    """
+
+    key: PUBLIC_KEY
+    """
+    A string representation of the Auxiliary public key.  
+    It is up to the external `AuxiliaryEncrypt` function to know how to parse this value
+    """
+
+
+AuxiliaryEncrypt = Callable[[MESSAGE, PUBLIC_KEY], Optional[ENCRYPTED_MESSAGE]]
+"""A callable type that represents the auxiliary encryption scheme."""
+
+AuxiliaryDecrypt = Callable[[ENCRYPTED_MESSAGE, SECRET_KEY], Optional[MESSAGE]]
+"""A callable type that represents the auxiliary decryption scheme."""
--- a/src/electionguard/rsa.py
+++ b/src/electionguard/rsa.py
@@ -1,0 +1,108 @@
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.asymmetric.rsa import (
+    generate_private_key,
+    RSAPrivateKey,
+    RSAPrivateKeyWithSerialization,
+    RSAPublicKey,
+)
+from cryptography.hazmat.primitives.asymmetric.padding import PKCS1v15
+from cryptography.hazmat.primitives.serialization import (
+    Encoding,
+    load_pem_private_key,
+    load_pem_public_key,
+    NoEncryption,
+    PrivateFormat,
+    PublicFormat,
+)
+from typing import Optional, NamedTuple
+
+
+PUBLIC_EXPONENT = 65537
+KEY_SIZE = 4096
+PADDING = 11
+MAX_BITS = int(KEY_SIZE / 8) - PADDING
+ISO_ENCODING = "ISO-8859-1"
+BYTE_ORDER = "big"
+
+
+class RSAKeyPair(NamedTuple):
+    """Key pair of RSA pkcs1 in bytes"""
+
+    private_key: str
+    public_key: str
+
+
+def rsa_keypair() -> RSAKeyPair:
+    """
+    Create RSA keypair
+
+    :return: RSA key pair
+    """
+
+    private_key: RSAPrivateKeyWithSerialization = generate_private_key(
+        public_exponent=PUBLIC_EXPONENT, key_size=KEY_SIZE, backend=default_backend()
+    )
+    private_key_bytes = private_key.private_bytes(
+        encoding=Encoding.PEM,
+        format=PrivateFormat.TraditionalOpenSSL,
+        encryption_algorithm=NoEncryption(),
+    )
+    private_key_string = str(private_key_bytes, ISO_ENCODING)
+
+    public_key: RSAPublicKey = private_key.public_key()
+    public_key_bytes = public_key.public_bytes(
+        encoding=Encoding.PEM, format=PublicFormat.SubjectPublicKeyInfo
+    )
+    public_key_string = str(public_key_bytes, ISO_ENCODING)
+
+    return RSAKeyPair(private_key_string, public_key_string)
+
+
+def rsa_encrypt(message: str, public_key: str) -> Optional[str]:
+    """
+    Encrypt with RSA public key
+
+    :param message: Message
+    :param key_pair: RSA public key
+    :return: Encrypted message
+    """
+    data = bytes(public_key, ISO_ENCODING)
+    rsa_public_key: RSAPublicKey = load_pem_public_key(data, backend=default_backend())
+    integer = int(message)
+    bits = count_set_bits(integer)
+    if bits > MAX_BITS:
+        return None
+    plaintext = integer.to_bytes(bits, BYTE_ORDER)
+    ciphertext = rsa_public_key.encrypt(plaintext, PKCS1v15())
+    return str(ciphertext, ISO_ENCODING)
+
+
+def rsa_decrypt(encrypted_message: str, private_key: str) -> Optional[str]:
+    """
+    Decrypt with RSA private key
+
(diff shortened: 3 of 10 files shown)


Pipfile
```diff
@@ -7,18 +7,19 @@ verify_ssl = true
 black = "*"
 coverage = "*"
 docutils = "*"
+hypothesis = "==5.15.1"
```
keithrfung, line 10: Moved these because the package is not dependent on these.

Pipfile
```diff
@@ -7,18 +7,19 @@ verify_ssl = true
...
 pylint = "*"
 pytest = "*"
+typish = '*'
 
 [packages]
-hypothesis = "==5.15.1"
 numpy = '==1.18.2'
 jsons = '==1.1.2'
-typish = '*'
+rsa = "*"
```
keithrfung, line 22: There were many other package options. This one was selected because it is a plain name, Apache 2.0, and is a very small library only focused on RSA.

src/electionguard/types.py
```diff
@@ -1,4 +1,9 @@
+from typing import TypeVar
+
 BALLOT_ID = str
 CONTEST_ID = str
 GUARDIAN_ID = str
 SELECTION_ID = str
+ENCRYPTION_OUTPUT = TypeVar("ENCRYPTION_OUTPUT")
+SECRET_KEY = TypeVar("SECRET_KEY")
+PUBLIC_KEY = TypeVar("PUBLIC_KEY")
```
keithrfung, line 9: Generic types were used to ensure the Auxiliary keys would be flexible. The hope is to allow flexibility.

src/electionguard/rsa.py
```diff
@@ -0,0 +1,89 @@
...
+    load_pem_public_key,
+    NoEncryption,
+    PrivateFormat,
+    PublicFormat,
+)
+from typing import NamedTuple
+
+
+PUBLIC_EXPONENT = 65537
+KEY_SIZE = 4096
```
keithrfung, line 21: 🚨🚨🚨🚨🚨ALERT🚨🚨🚨🚨🚨🚨

This seems to be related to the error in the build. As far as I can tell, there are test cases which exceed the 4096 bit character and the error is consistent.

src/electionguard/rsa.py
```diff
@@ -0,0 +1,89 @@
...
+    :param private_key: RSA private key
+    :return: Message
+    """
+
+    data = bytes(private_key, UTF_ENCODING)
+    rsa_private_key: RSAPrivateKey = load_pem_private_key(
+        data, password=None, backend=default_backend()
+    )
+    ciphertext = bytes(encrypted_message, ISO_ENCODING)
+    plaintext = rsa_private_key.decrypt(ciphertext, PKCS1v15())
```
keithrfung, line 88: This is where the padding is set. It is currently PKCS1v15()

https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#module-cryptography.hazmat.primitives.asymmetric.padding

src/electionguard/rsa.py
keithrfung: I have tried up to 8000 so I'm not quite sure where to go from here.

tests/test_decryption_mediator.py
```diff
@@ -205,6 +205,9 @@ def test_announce(self):
         # Cannot get plaintext tally without a quorum
         self.assertIsNone(subject.get_plaintext_tally())
 
+    def tearDown(self):
+        self.key_ceremony.reset(CeremonyDetails(self.NUMBER_OF_GUARDIANS, self.QUORUM))
+
```
keithrfung, line 210: This caused an incredible amount of stress.

keithrfung:
When the CI passes for a super frustrating PR,

Pipfile
```diff
@@ -7,18 +7,19 @@ verify_ssl = true
...
 pylint = "*"
 pytest = "*"
+typish = '*'
 
 [packages]
-hypothesis = "==5.15.1"
 numpy = '==1.18.2'
 jsons = '==1.1.2'
-typish = '*'
+cryptography = "*"
```
keithrfung, line 22: This is the RSA library being used.

src/electionguard/auxiliary.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    key: str
+    """
+    A string representation of the Auxiliary public key.  
+    It is up to the external `AuxiliaryEncrypt` function to know how to parse this value
+    """
+
+
+AuxiliaryEncrypt = Callable[[str, str], Optional[str]]
```
keithrfung, line 34: These functions now allow None to allow for errors

src/electionguard/rsa.py
```diff
@@ -0,0 +1,108 @@
...
+
+    :param message: Message
+    :param key_pair: RSA public key
+    :return: Encrypted message
+    """
+    data = bytes(public_key, ISO_ENCODING)
+    rsa_public_key: RSAPublicKey = load_pem_public_key(data, backend=default_backend())
+    integer = int(message)
+    bits = count_set_bits(integer)
+    if bits > MAX_BITS:
```
keithrfung, line 73: This is a check to ensure that we don't try to encrypt a larger entity than our key_size allows.

keithrfung:
There should probably be some discussion on ensuring tests bypass the rsa check when possible in certain sections. There are particularly sections where this adds a considerable amount of time to the CI build. We could use the override in these particular sections and just return the message itself to ensure a faster build time. 

@AddressXception Perhaps this should be an issue? Or is this relevant to this PR?

src/electionguard/auxiliary.py
```diff
@@ -0,0 +1,43 @@
...
+
+from .types import GUARDIAN_ID
+
+MESSAGE = str
+PUBLIC_KEY = str
+SECRET_KEY = str
+ENCRYPTED_MESSAGE = str
+
+
+class AuxiliaryKeyPair(NamedTuple):
```
keithrfung, line 11: Created a new file for this to isolate this for future refactoring.

src/electionguard/auxiliary.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+
+    key: PUBLIC_KEY
+    """
+    A string representation of the Auxiliary public key.  
+    It is up to the external `AuxiliaryEncrypt` function to know how to parse this value
+    """
+
+
+AuxiliaryEncrypt = Callable[[MESSAGE, PUBLIC_KEY], Optional[ENCRYPTED_MESSAGE]]
```
keithrfung, line 39: This now returns Optional to allow for handling errors similar to the rest of the code base. I'm not particularly in favor of this but it's consistency so it remains. 

The `str` type was given names to help new users identify exactly how these functions worked.

keithrfung:
Added this issue regarding the slower CI times --> #103

PR #104, opened 2020-07-14
Remove RSA from certain tests and CI for speed

### LINKED ISSUE #103: Override default aux encrypt/decrypt for test and CI builds
Reduce the continuous integration build time for PR's by overriding any test that include the rsa encryption except for the direct unit test (test_rsa.py) and integration test. This means adding a AuxiliaryEncrypt and AuxiliaryDecrypt. A simple solution is probably taking the key but always return the same message. 

```py
def fake_encrypt(message: str, public_key: str) -> Optional[str]:
      return message 
```

keithrfung:
### Issue

Fixes #103 

### Description
- Add identity encrypt / decrypt methods
- update methods to allow encrypt/decrypt override at higher levels
- update tests to use identity for speed

### Testing
- run full test suite

### Checklist

💚Thank you!

### DIFF
--- a/src/electionguard/decryption.py
+++ b/src/electionguard/decryption.py
@@ -2,6 +2,7 @@ from multiprocessing import cpu_count
 from multiprocessing.dummy import Pool
 from typing import Dict, Optional, Union
 
+from .auxiliary import AuxiliaryDecrypt
 from .ballot import CiphertextAcceptedBallot, CiphertextBallotSelection
 from .data_store import DataStore
 from .decryption_share import (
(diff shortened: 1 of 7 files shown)


src/electionguard/auxiliary.py
```diff
@@ -41,3 +42,33 @@ class AuxiliaryPublicKey(NamedTuple):
 
 AuxiliaryDecrypt = Callable[[ENCRYPTED_MESSAGE, SECRET_KEY], Optional[MESSAGE]]
 """A callable type that represents the auxiliary decryption scheme."""
+
+
+def identity_auxiliary_encrypt(message: str, public_key: str) -> Optional[str]:
```
AddressXception, line 47: i dont think we should include this in the core lib

keithrfung: Ok. I'll manually make lambdas in the functions.

PR #113, opened 2020-07-20
Deserialization bugfixes

danwallach:
### Issue
TBD
### Description
Deserialization hits a variety of bugs, some in ElectionGuard, some in the `jsons` library. This work-in-progress PR tries to address these concerns.

### Testing
Relevant end-to-end unit tests are in the `arlo-e2e` project. Will port them over here if necessary. TBD

### Checklist

💚Thank you!


Pipfile
```diff
@@ -19,7 +19,7 @@ typish = '*'
 [packages]
 numpy = '==1.18.2'
 jsons = '==1.1.2'
-cryptography = "*"
+cryptography = "==2.9.2"
```
keithrfung, line 22: This should be an isolated issue and PR if there needs to be a locked version on the cryptography for RSA.

src/electionguard/publish.py
```diff
@@ -1,20 +1,19 @@
...
-from .group import ElementModP, ElementModQ
 
-RESULTS_DIR = "results" + path.sep
-COEFFICIENTS_DIR = RESULTS_DIR + "coefficients" + path.sep
-BALLOTS_DIR = RESULTS_DIR + "encrypted_ballots" + path.sep
-SPOILED_DIR = RESULTS_DIR + "spoiled_ballots" + path.sep
+RESULTS_DIR = "results"
+COEFFICIENTS_DIR = path.join(RESULTS_DIR, "coefficients")
+BALLOTS_DIR = path.join(RESULTS_DIR, "encrypted_ballots")
+SPOILED_DIR = path.join(RESULTS_DIR, "spoiled_ballots")
```
keithrfung, line 16: The `path.join` is a good addition but this should likely be an isolated issue and PR since it pertains to the export / serialization of the data and not the import / deserialization.

danwallach: It impacted me because I wasn't putting path separators in my own code.

keithrfung: That may be true. It still should be isolated as separate PR.

src/electionguard/ballot.py
```diff
@@ -573,9 +628,9 @@ def hashed_ballot_nonce(self) -> Optional[ElementModQ]:
         """
 
         if self.nonce is None:
-            log_warning(
-                f"missing nonce for ballot {self.object_id} could not derive from null nonce"
-            )
+            # log_warning(
```
keithrfung, line 631: Why is this commented out?

Pipfile
danwallach: This probably isn't the problem. The issue was that ElectionGuard wasn't declaring `cryptography` as a dependency, so when I imported it, it didn't recursively pull `cryptography` in. I don't know enough about how Python distribution package are built. Somewhere you need to  declare this. Probably other packages as well.

src/electionguard/ballot.py
danwallach: Because it's generating a warning when the nonce is None, and in fact, it's a completely normal use case for the nonce to be None. That's how every CiphertextAcceptedBallot is.

src/electionguard/publish.py
danwallach: Okay, I'm fine with that. FWIW, I'm depending on this behavior now.

src/electionguard/ballot.py
keithrfung: If that's the case, it should be removed not commented. There is no value in commented code.

Pipfile
keithrfung: Created this issue for this: #115

src/electionguard/publish.py
keithrfung: Created issue #116

keithrfung:
@danwallach 

I isolated the key components of this PR into separate pull requests and corresponding issues. 
The large one with key fixes surrounding deserialization is #120 -> PR #119 
I still have to fix all the failing tests since those aren't resolved. 

I'm going to close this Draft and work can continue from that point. I've rolled all the individual changes from PRs and some of your particular arlo specific changes into `temp/arlo-deserialization-patch`. This will make merging later and changing back to `master` a bit easier.

**Note: Branching naming**
In the future make sure you name your branches that are directly off the repo with `docs/`, `feature/`, `fix/`, or `temp/` prefixes. 
Screen Shot 2020-07-21 at 4 50 11 PM

PR #114, opened 2020-07-21
Refactor to optimize output data for readability / usability

### LINKED ISSUE #90: Refactor algebraic variables to descriptive names
Refactor existing variables in equations and proofs to have a more meaningful descriptive name. Examples would be `a`, `b`, `p` and `q`. 

The refactor of each name will correspond to the next specification being worked on. The docstrings and docs should be updated accordingly still mention the original variable when renamed. For example:
Old: 
```py
a
```

New: 
```py
meaningful_name
"""Meaningful_thing also known as a"""
```

r is cofactor
g is generator
p is large prime
q is small prime

on elgamal ciphertext:
alpha rename to pad
beta rename to data

ciphertect ballot selections
- rename the elgamal message field to "encrypted vote" 

plaintext tally
- the plaintext field rename to tally
- the tally field rename to exponential tally

### LINKED ISSUE #91: Remove the inverse of g as a constant
The inverse of g is no longer as present in the existing proofs so it may be removed to shrink code size.

### LINKED ISSUE #92: Refactor proof names to be more descriptive
Most proofs are labeled after their originating mathematical functions. This isn't as useful to developers and new users. 

The classes themselves should be renamed to items like `SelectionLimitProof`
There should be a type added under these that define a type as an enum. This can include the formal name and the descriptive name. This is useful for output.
The docstrings should detail the mathematical function being used for clarity purposes.

### LINKED ISSUE #93: Refactor description fields for results output
This is to rename fields on the results output so that individuals can better understand the resulting output. These changes will be changed on the classes themselves and are still in flux. Additions should be added as comments here.

keithrfung:
### Issue

Fixes #90  
Fixes #91 
Fixes #92
Fixes #93 

### Description

This pull request is going to contain several issues. This is not a recommended approach but due to the sweeping changes across classes and the dependency of the issues This is an exception. The recommendation will also be a **rebase** of this PR instead of a squash. The goal of this PR is to provide better output format and to ensure more information for individuals using just the raw data. 

**Inverse of G**
Inverse of G was originally kept in to be a constant of the election but since G is provided, this was seen as excess data. This was removed to be replaced by the calculation in the single test location.

**Renaming proofs**
The proofs are exported in json with very little information. The goal is not to just have the named items in the tuple but a little more information. The removal of `NamedTuple`s to frozen `dataclasses` became necessary to include the string data without complicated constructor sequence. An alternate version was considered with a NamedTuple that just included the usage and the name was drawn from the serialization. 

**Rename algebraic variables**
Refactor Elgamel Ciphertext, Constants, Ballots, and Tallies to use more understandable language on outputs for non technical users. 

### Testing
- This is best checked by running the full test suite. 

### Checklist

💚Thank you!


src/electionguard/ballot.py
```diff
@@ -165,8 +165,8 @@ class CiphertextBallotSelection(ElectionObjectBase, CryptoHashCheckable):
     # The SelectionDescription hash
     description_hash: ElementModQ
 
-    # The encrypted representation of the plaintext field
-    message: ElGamalCiphertext
+    # The encrypted representation of the vote field
+    encrypted_data: ElGamalCiphertext
```
keithrfung, line 169: A hidden issue here is `CiphertextSelection` aka a union class above this level. It forces both the `CiphertextBallotSelection` and `CiphertextTallySelection` to have the encrypted text use the same variable name. It is not called out and I think it's slightly dangerous for future refactors.

Maybe refactoring them both to inherit from an abstract class with the singular `encrypted_data` field would be better? With the use of dataclass this would change all the parameter ordering above it.

src/electionguardtest/data/ballot_in_simple.json
```diff
@@ -7,11 +7,11 @@
             "ballotSelections": [
                 {
                     "objectId": "john-adams-selection",
-                    "plaintext": "True"
+                    "vote": "True"
```
keithrfung, line 10: The renaming in this file is going to collide with some of the other PR's

src/electionguard/ballot.py
AddressXception: we could use a protocol instead to enforce the naming convention on any object type

src/electionguard/ballot.py
```diff
@@ -2,7 +2,7 @@
 from datetime import datetime
 from enum import Enum
 from distutils import util
-from typing import Optional, List, Any, Sequence
+from typing import Any, List, Optional, Protocol, runtime_checkable, Sequence
```
AddressXception, line 6: in this file in the `CiphertextBallot` we need to pass in the `seed_hash` and cache it on the object for serialization.  While we are in there, we should make 2 more changes:
1. `generate_tracking_id` should be a class-level function and a constructor input, or a similar appropriate pattern so the object can be made immutably
2. we should change all of the `# comments` on the object fields in this file to `"""summary comments"""` so they show up in intellisense

src/electionguard/ballot.py
```diff
@@ -612,31 +609,34 @@ class CiphertextBallot(ElectionObjectBase, CryptoHashCheckable):
...
     contests: List[CiphertextBallotContest]
+    """List of contests for this ballot"""
 
-    # the unique ballot tracking id for this ballot
     tracking_id: Optional[ElementModQ] = field(init=False)
+    """Unique ballot tracking id for this ballot"""
 
-    # timestamp in ticks
     timestamp: int = field(init=False)
+    """Timestamp in ticks"""
```
AddressXception, line 628: """ Timestamp at which the ballot encryption is generated in tick"""?

keithrfung: Got it. working on it.

PR #119, opened 2020-07-21
🗳 Deserialization with Ballots

### LINKED ISSUE #120: Deserializing Ballots
The ballots and underlying classes have issues being deserialized stemming from issues with `jsons` library casing and use of `InitVar`. 

Reported by @danwallach in draft #113

keithrfung:
### Issue

Fixes #120 

### Description
- Add make_ methods to reduce constructor issues
- Remove InitVar
- Add deserialization for ElementModP and ElementModQ

### Testing
- attempt to deserialize published files

### Checklist

💚Thank you!

### DIFF
--- a/src/electionguard/ballot.py
+++ b/src/electionguard/ballot.py
@@ -1,7 +1,7 @@
-from dataclasses import dataclass, field, InitVar
+from dataclasses import dataclass, field
 from datetime import datetime
-from enum import Enum
 from distutils import util
+from enum import Enum
 from typing import Optional, List, Any, Sequence
 
 from .chaum_pedersen import (
@@ -16,7 +16,7 @@ from .group import add_q, ElementModP, ElementModQ, ZERO_MOD_Q
 from .hash import CryptoHashCheckable, hash_elems
 from .logs import log_warning
 from .tracker import get_rotating_tracker_hash, tracker_hash_to_words
-from .utils import to_ticks
+from .utils import to_ticks, flatmap_optional
 
 
 def _list_eq(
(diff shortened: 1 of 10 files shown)


src/electionguard/ballot.py
```diff
@@ -261,8 +232,59 @@ def crypto_hash_with(self, seed_hash: ElementModQ) -> ElementModQ:
...
+    nonce: Optional[ElementModQ] = None,
+    crypto_hash: Optional[ElementModQ] = None,
+    proof: Optional[DisjunctiveChaumPedersenProof] = None,
+    extended_data: Optional[ElGamalCiphertext] = None,
+) -> CiphertextBallotSelection:
+    """
+    Constructs a `CipherTextBallotSelection` object. Most of the parameters here match up to fields
+    in the class, but this helper function will optionally compute a Chaum-Pedersen proof if the
+    given nonce isn't `None`. Likewise, if a crypto_hash is not provided, it will be derived from
+    the other fields.
```
AddressXception, line 263: i think we should call out each parameter in the summary description since there are so many

danwallach: I want to emphasize this. I'm still only 50% sure what the different sorts of hashes are. So long as you're trying to make these things more clear, coming up with better nomenclature for the various hashes floating around the system sounds really important.

keithrfung: The better naming on the hashes was something also mentioned by Josh about the output data and should probably be in PR #114 . The issue is that there weren't a lot of good names for these in the first place.

keithrfung:
In relations to this PR, we're going to move the serialization and deserialization setup into serializable by moving the imports into the method. Without this, we end up with a circular dependency. However, we want to really restrict `jsons` to a single file.

PR #123, opened 2020-07-24
added check to suppress unnecessary log warning

### LINKED ISSUE #122: Unnecessary / incorrect log warning when serializing a CiphertextAcceptedBallot
The warning, in `ballot.py`, around or about line 630:
```
        if self.nonce is None:
            log_warning(
                f"missing nonce for ballot {self.object_id} could not derive from null nonce"
            )
```
This warning is only the right thing to do if you're dealing with a regular `CiphertextBallot`, but not a `CiphertextAcceptedBallot`, which by definition no longer has the nonce used for the encryption. Since serializing a list of these things will potentially generate a large stack of these warnings, the right thing to do is to suppress the warning in this case.

danwallach:
### Issue

Fixes #122 

### Description
Makes the warning conditional on the instance type of the CiphertextBallot. If it's a CiphertextAcceptedBallot, then no warning is generated. Otherwise, behavior is identical.

(This patch is only relevant to the `fix/deserialization-issues` branch, where it's -3+4 lines. If you compare it to `main`, it's much bigger.)

### Testing
In regular use, should have no effect on testing. But logs will be cleaner, easier to read.

### Checklist

💚Thank you!

### DIFF
--- a/src/electionguard/ballot.py
+++ b/src/electionguard/ballot.py
@@ -626,14 +626,6 @@ class CiphertextBallot(ElectionObjectBase, CryptoHashCheckable):
         :return: a hash value derived from the description hash, the object id, and the nonce value
                 suitable for deriving other nonce values on the ballot
         """
-
-        if self.nonce is None:
-            if not isinstance(self, CiphertextAcceptedBallot):
-                log_warning(
-                    f"missing nonce for ballot {self.object_id} could not derive from null nonce"
-                )
-            return None
-
         return hash_elems(self.description_hash, self.object_id, self.nonce)
 
     def generate_tracking_id(self, seed_hash: ElementModQ) -> None:
--- a/src/electionguard/ballot.py
+++ b/src/electionguard/ballot.py
@@ -628,9 +628,10 @@ class CiphertextBallot(ElectionObjectBase, CryptoHashCheckable):
         """
 
         if self.nonce is None:
-            log_warning(
-                f"missing nonce for ballot {self.object_id} could not derive from null nonce"
-            )
+            if not isinstance(self, CiphertextAcceptedBallot):
+                log_warning(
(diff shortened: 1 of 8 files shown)


src/electionguard/ballot.py
```diff
@@ -628,9 +628,10 @@ def hashed_ballot_nonce(self) -> Optional[ElementModQ]:
         """
 
         if self.nonce is None:
-            log_warning(
-                f"missing nonce for ballot {self.object_id} could not derive from null nonce"
-            )
+            if not isinstance(self, CiphertextAcceptedBallot):
```
AddressXception, line 631: why dont we just get rid of the error message entirely? or change it to `log_debug`?  i think that's preferable over checking the subclassed type, right?  Open/Close Principle?

danwallach: I'm fine with nuking this entirely. If you're going to keep any sort of logging here, it should only happen with dealing with CiphertextBallots in other states. A CiphertextAcceptedBallot will never have a nonce in it.

danwallach:
So I had an a-ha moment. This particular method doesn't actually care if the nonce is absent. We're computing a hash value over various things. If the nonce isn't there, we can still compute a hash value. `hash_elems` doesn't care. The only issue is whether we want this particular hash value to be the same for a `CiphertextBallot` (with nonce) and an otherwise equal `CiphertextAcceptedBallot` (without nonce).

Anyway, I pushed a change that gets rid of the warning check altogether. I think this is fine.

danwallach:
Food for thought: should the nonce actually be included in this particular computation? I don't know exactly where we're calling this method, but if we want the same result for a CiphertextBallot and its corresopnding CiphertextAcceptedBallot, then the nonce should not be included in what's being passed to `hash_elems`.

PR #127, opened 2020-07-28
✨ Implement Scheduler for multi processing

### LINKED ISSUE #54: Consistent Multiprocessing for Tally
multiprocessing is used in a few places to speed up computation of some large values.  When using the library in the recommended way, it behaves as expected, however it is possible to deadlock by calling internal functions.  This is illustrated by `Pool` objects existing in multiple scopes.

Proposed solution:

Implement a consistent queuing strategy for scheduling work on multiple processors so that the library cannot be deadlocked intentionally or unintentionally.

The focus on this card is going to be moved to displaying an example of paralleling for Tallies specifically.

keithrfung:
### Issue

Fixes #54 

### Description
- Add scheduler singleton to handle creating pools an closing pools. 
- Remove Pool references and replace scheduler
- Update tests and benchmarks

### Testing
- Full test suite but particularly encryption / decryption

### Checklist

💚Thank you!

### DIFF
--- a/bench/bench_chaum_pedersen.py
+++ b/bench/bench_chaum_pedersen.py
@@ -1,6 +1,5 @@
-from multiprocessing import Pool, cpu_count
 from timeit import default_timer as timer
-from typing import Tuple, NamedTuple, Dict
+from typing import Dict, List, NamedTuple, Tuple
 
 from numpy import average, std
 
@@ -13,7 +12,7 @@ from electionguard.elgamal import (
 )
 from electionguard.group import ElementModQ, int_to_q_unchecked
 from electionguard.nonces import Nonces
-
+from electionguard.scheduler import Scheduler
 from electionguard.utils import get_optional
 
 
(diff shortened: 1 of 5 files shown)


danwallach:
FWIW, I'm not sure this is going to fix the performance problems. When I wrote the equivalent code in `arlo-e2e` (see, e.g., https://github.com/votingworks/arlo-e2e/blob/25c59801fccaa78352cb10cdd7478e36a178cdb0/src/arlo_e2e/tally.py#L82), I was very careful about the granularity of the tasks being performed in parallel.

Encrypting ballots: each ballot is one task, and I map the encryption function over a list of ballots. 

Tallying ballots: I rearrange the ballots based on the selection `object_id` values, and then the parallelism maps over the selection's ids (so, one core per column of data). 

Also, for what it's worth, I love using `tqdm` (and there are ten other libraries like it), which you can wrap around any iterable, and it then prints a progress meter. Super handy.

keithrfung:
@danwallach Thanks. We're already aware there are little performance gains as is. This PR also is to minimize the dependency of pool spread over multiple places, but we're already shifting it to draft and adjusting it. 

@AddressXception is already building on this PR and adjusting the tally to exemplify a parallel workflow pattern.

PR #131, opened 2020-08-03
Fix/discrete log async

### LINKED ISSUE #130: discrete_log initialization doesn't work if there isn't an "event loop"
I'm working on getting EG to work with `ray.io`, a cluster computing framework. I ran into one issue with `async` code in `dlog.py`. The gist of it is that the call to initialize `__dlog_lock` was failing, and I was getting an error about a missing "event loop". Turns out, the way Ray was running the code was tickling this issue.

The solution is really straightforward: delay initialization of the lock variable until you're inside `__discrete_log_internal`.

danwallach:
### Issue
Fixes #130 

### Description
I'm working on getting EG to work with ray.io, a cluster computing framework. I ran into one issue with async code in dlog.py. The gist of it is that the call to initialize __dlog_lock was failing, and I was getting an error about a missing "event loop". Turns out, the way Ray was running the code was tickling this issue.

The solution is really straightforward: delay initialization of the lock variable until you're inside __discrete_log_internal.

### Testing
The existing test suite still works fine.

### Checklist

💚Thank you!

### DIFF
--- a/src/electionguard/dlog.py
+++ b/src/electionguard/dlog.py
@@ -9,7 +9,6 @@ __dlog_cache: Dict[ElementModP, int] = {ONE_MOD_P: 0}
 __dlog_max_elem = ONE_MOD_P
 __dlog_max_exp = 0
 
-# initialized inside __discrete_log_internal
 __dlog_lock: Optional[asyncio.Lock] = None
 
 
@@ -41,15 +40,7 @@ async def __discrete_log_internal(e: ElementModP) -> int:
     global __dlog_lock
 
     if __dlog_lock is None:
-        # We cannot run asyncio.Lock() if we don't have an "event loop", which
-        # can happen if we're running in certain environments. The solution
-        # is to put this bit of initialization into an async function, which
-        # is where we happen to be right now. With Python's relatively simple
-        # concurrency model, we don't have to worry about multiple threads
-        # arriving here simultaneously. This code will run exactly once. If
-        # you're using one of the many multiprocessing libraries, this will
-        # run exactly once per process.
-
+        # Initialize the lock on on first function call per process
         __dlog_lock = asyncio.Lock()
 
     async with __dlog_lock:
--- a/src/electionguard/dlog.py
+++ b/src/electionguard/dlog.py
@@ -1,14 +1,16 @@
 # support for computing discrete logs, with a cache so they're never recomputed
 
 import asyncio
-from typing import Dict
+from typing import Dict, Optional
 
 from .group import G, ElementModP, ONE_MOD_P, mult_p, int_to_p_unchecked
 
 __dlog_cache: Dict[ElementModP, int] = {ONE_MOD_P: 0}
 __dlog_max_elem = ONE_MOD_P
 __dlog_max_exp = 0
-__dlog_lock = asyncio.Lock()
+
+# initialized inside __discrete_log_internal
+__dlog_lock: Optional[asyncio.Lock] = None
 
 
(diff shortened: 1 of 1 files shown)


danwallach:
Need to rebase this before merge. The change here is way too big.

danwallach:
Okay, rebase successful. The PR now only touches `dlog.py`.

src/electionguard/dlog.py
```diff
@@ -38,6 +40,18 @@ async def __discrete_log_internal(e: ElementModP) -> int:
     global __dlog_max_exp
     global __dlog_lock
 
+    if __dlog_lock is None:
+        # We cannot run asyncio.Lock() if we don't have an "event loop", which
```
keithrfung, line 44: This is the description that should be on the PR, not in code. 

Perhaps add a bit more information on the method description.

src/electionguard/dlog.py
```diff
@@ -1,14 +1,16 @@
...
+from typing import Dict, Optional
 
 from .group import G, ElementModP, ONE_MOD_P, mult_p, int_to_p_unchecked
 
 __dlog_cache: Dict[ElementModP, int] = {ONE_MOD_P: 0}
 __dlog_max_elem = ONE_MOD_P
 __dlog_max_exp = 0
-__dlog_lock = asyncio.Lock()
+
+# initialized inside __discrete_log_internal
```
keithrfung, line 12: Unneeded comment.

src/electionguard/dlog.py
danwallach: I guess we disagree about how comments can be useful to future readers of the code. This code is incredibly subtle, depending on details of how Python's lack of a multithreaded coding model works. In fact, if I really understand it, we could nuke all the asynchrony altogether, since Python never has two threads running concurrently. I'm a little hesitant to do that, so at least I can leave behind an artifact of my thinking for whoever tackles this next.

AddressXception: i modified this. ill explain in a comment in my diff

src/electionguard/dlog.py
```diff
@@ -38,6 +39,10 @@ async def __discrete_log_internal(e: ElementModP) -> int:
     global __dlog_max_exp
     global __dlog_lock
 
+    if __dlog_lock is None:
+        # Initialize the lock on on first function call per process
```
AddressXception, line 43: We cannot run asyncio.Lock() if we don't have an "event loop", which can happen if we're running in certain environments. The solution is to put this bit of initialization into an async function, which is where we happen to be right now. With Python's relatively simple concurrency model, we don't have to worry about multiple threads arriving here simultaneously. This code will run exactly once. If you're using one of the many multiprocessing libraries, this will run exactly once per process.

src/electionguard/dlog.py
```diff
@@ -38,6 +39,10 @@ async def __discrete_log_internal(e: ElementModP) -> int:
...
+        __dlog_lock = asyncio.Lock()
```
AddressXception, line 44: We cannot run asyncio.Lock() if we don't have an "event loop", which can happen if we're running in certain environments. The solution is to put this bit of initialization into an async function, which is where we happen to be right now. With Python's relatively simple concurrency model, we don't have to worry about multiple threads arriving here simultaneously. This code will run exactly once. If you're using one of the many multiprocessing libraries, this will run exactly once per process.

AddressXception:
modified the comment lines to be more concise in code, and moved the description into the PR.  Everything is ready to go

PR #134, opened 2020-08-03
Feature/multi processing

### LINKED ISSUE #54: Consistent Multiprocessing for Tally

AddressXception:
### Issue
*Link your PR to an issue*

Fixes #54 

### Description

Add a pattern for handling multiprocessing by creating a singleton scheduler class that wraps the python multiprocessing library

### Testing

Run the tests

### Checklist

💚Thank you!


bench/bench_chaum_pedersen.py
```diff
@@ -49,62 +48,64 @@ def identity(x: int) -> int:
...
-        print(f"    Stddev = {std_verify_scalar:.6f} sec")
-
-        # Run in parallel
-        start_all_parallel = timer()
-        timing_data_parallel = pool.map(chaum_pedersen_bench, inputs)
-        end_all_parallel = timer()
-
-        speedup[size] = (end_all_scalar - start_all_scalar) / (
-            end_all_parallel - start_all_parallel
+    with Scheduler() as scheduler:
```
AddressXception, line 53: this is one way to use the scheduler.  it will automatically be closed at the end of the loop

src/electionguard/decryption.py
```diff
@@ -111,17 +110,20 @@ def compute_decryption_share_for_cast_contests(
...
-        selection_decryptions = cpu_pool.starmap(
+        selection_decryptions: List[
+            Optional[CiphertextDecryptionSelection]
+        ] = scheduler.schedule(
             compute_decryption_share_for_selection,
             [
                 (guardian, selection, context)
                 for (_, selection) in contest.tally_selections.items()
             ],
+            with_shared_resources=True,
```
AddressXception, line 126: the `with_shared_resources` flag tells the scheduler to schedule in a way that shares state across the different iterations

src/electionguard/scheduler.py
```diff
@@ -0,0 +1,103 @@
...
+from psutil import cpu_count
+import traceback
+from typing import Any, Callable, Iterable, List
+
+from .logs import log_warning
+from .singleton import Singleton
+from .utils import T
+
+
+class Scheduler(Singleton, AbstractContextManager):
```
AddressXception, line 15: this implementation is generic enough that it should be straightforward to override this pattern and implement multiprocessing in some other way - such as via a cluster.  in the future we could turn this into a protocol and allow for alternative implementations but that is beyond the scope of this work

src/electionguard/scheduler.py
```diff
@@ -0,0 +1,103 @@
...
+        self.__thread_pool = ThreadPool(max_processes)
+
+    def close(self) -> None:
+        """Close pools"""
+        self.__process_pool.close()
+        self.__thread_pool.close()
+
+    def cpu_count(self) -> int:
+        """Get CPU count"""
+        return cpu_count(logical=False)
```
AddressXception, line 54: using this version of cpu_count allows us to specify only using physical cores.  Adding logical cores doesn't really get us anything for the work we are trying to schedule on most CPU's - and in fact the overhead can actually decrease performance depending on the CPU architecture and what data is flowing into the caches

src/electionguard/tally.py
```diff
@@ -90,13 +90,24 @@ class CiphertextTallyContest(ElectionObjectBase):
     A collection of CiphertextTallySelection mapped by SelectionDescription.object_id
     """
 
+    # TODO: rename accumulate ballot
```
AddressXception, line 93: TODO: fiox or remove this

danwallach:
FWIW, I'm in the middle of doing a `ray.io` port, and it's just not at all the same as `multiprocessing`. I'm just tearing up and rebuilding things to work with Ray. 

My advice: have a "top level function" that runs the various phases of encryption / tallying / etc., where that function does all the parallel dispatch, and where that function just needs to be rewritten to deal with different parallel libraries.

Makefile
```diff
@@ -8,6 +8,9 @@ IS_64_BIT ?= $(shell python -c 'from sys import maxsize; print(maxsize > 2**32)'
 
 all: environment install validate lint coverage
 
+bench:
+	pipenv run python -s bench/bench_chaum_pedersen.py
```
keithrfung, line 12: We should really run all these in the folder, but I think it's simpler to keep adding lines like this for cleanliness.

I'd add `@echo 📊 BENCHMARKS` ahead of this.

setup.py
```diff
@@ -62,5 +62,11 @@
     classifiers=CLASSIFIERS,
     project_urls=PROJECT_URLS,
     python_requires="~=3.8",
-    install_requires=["gmpy2>=2.0.8", "numpy>=1.18.2", "jsons>=1.1.2", "cryptography",],
+    install_requires=[
+        "gmpy2>=2.0.8",
+        "numpy>=1.18.2",
+        "jsons>=1.1.2",
+        "cryptography",
+        "psutil>=5.7.2",
```
keithrfung, line 70: 👍 Good job, ensuring this was here.

src/electionguard/tally.py
keithrfung: Yeah. I'd be remiss if I didn't say we need to remove this or give it an issue.

tests/test_encrypt.py
```diff
@@ -263,7 +263,7 @@ def test_encrypt_simple_contest_referendum_succeeds(self):
         self.assertTrue(result.is_valid_encryption(hash_context, keypair.public_key))
 
     @settings(
-        deadline=timedelta(milliseconds=2000),
+        deadline=timedelta(milliseconds=4000),
```
keithrfung, line 266: Do these still go up in time?  Shouldn't these stay the same with the improvements?

tests/test_scheduler.py
```diff
@@ -0,0 +1,44 @@
...
+def _callable(data: int):
+    return data
+
+
+def _exception_callable(something: int):
+    raise Exception
+
+
+class TestScheduler(TestCase):
+    def test_schedule_callable_throws(self):
```
keithrfung, line 15: Nit-picky but we should stick to our arrange, act, assert comments.

src/electionguard/scheduler.py
```diff
@@ -0,0 +1,103 @@
...
+    ) -> List[T]:
+        """Safe wrapper around starmap to ensure pool is open"""
+        try:
+            return pool.starmap(task, arguments)
+        except ValueError as e:
+            log_warning(
+                f"safe_starmap({task}, {arguments}) exception ValueError({str(e)})"
+            )
+            return []
+        except:
```
keithrfung, line 86: These generic errors will get caught by LGTM. 

Any `except` or `BaseException` catches will give a warning.

keithrfung: https://lgtm.com/rules/6780080/

tests/test_encrypt.py
AddressXception: The Mac VM seems to have some variance on the amount of time it takes to run these tests.  i was getting false failures due to the timeout being too short so i increased them.

PR #139, opened 2020-08-05
✨ Fix deserialization issues

### LINKED ISSUE #120: Deserializing Ballots

### LINKED ISSUE #122: Unnecessary / incorrect log warning when serializing a CiphertextAcceptedBallot

### LINKED ISSUE #132: Deserialization issue with Proof Usage
Here's the code to deserialize a ProofUsage enum:
```
    set_deserializer(
        lambda usage_string, cls, **_: ProofUsage[usage_string], ProofUsage
    )
```

You can't just use `ProofUsage[usage_string]`. That crashes, as below:

```
>>> from electionguard.proof import ProofUsage
>>> ProofUsage["SecretValue"]
<ProofUsage.SecretValue: 'Prove knowledge of secret value'>
>>> ProofUsage["Prove knowledge of secret value"]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python@3.8/3.8.2/Frameworks/Python.framework/Versions/3.8/lib/python3.8/enum.py", line 344, in __getitem__
    return cls._member_map_[name]
KeyError: 'Prove knowledge of secret value'
```

You need to convert from the string (“Prove knowledge of secret value”) back to the name of the enum (“SecretValue”) before you can do the lookup like this.

To try to catch these and other issues, I wrote a new Hypothesis unit test in `test_serialization.py` and found several other issues along these lines, where deserialization was not the inverse of serialization.

keithrfung:
### Issue

Fixes #132 
Fixes #120 
Fixes #122 while obsoleting PR #123 

### Description

For many of the core ElectionGuard classes, deserialization of a serialized class does not yield the original or has issues.

**Replace post_init with factory methods**
Deserialization has issues with the post_init methods that we'd been using, similar to `InitVar`.  Factory `make_` methods were created for constructing the classes. 

**Resolve equality methods for testing**
To make sure equality testing worked among the core ElectionGuard classes, there were some additional changes. Some of them use @dataclass(eq=True...) to auto-generate the methods. This doesn't work for lists which might come back in a different order than they were written out. For those cases, existing _list_eq() method is used in custom __eq__ methods.

**Remove frozen tag**
A frozen=True dataclass with a non-frozen superclass causes an inheritance error (ignored by Python, flagged by PyCharm) to have . Since our Serializable isn't frozen, this is a problem. The frozen attribute was removed from the dataclasses only relevant to this PR.

**Change crypto_hash to class method from property**

**Add serialization methods for datetime**

**Allow strings to be passed into mpz for conversion**

**Resolve issues with Proof deserialization**
Proof serialized correctly but had issues deserializing with the conversions

**Add "None" deserialization**
"None" was not deserializing properly and caused issues with certain Optionals

### Testing

- `test_serializable`
- `test_end_to_end_election`

## Checklist

💚Thank you!


src/electionguard/tally.py
```diff
@@ -177,6 +177,20 @@ class PlaintextTally(ElectionObjectBase):
     spoiled_ballots: Dict[BALLOT_ID, Dict[CONTEST_ID, PlaintextTallyContest]]
 
 
+@dataclass
+class PublishedPlaintextTally(ElectionObjectBase):
```
keithrfung, line 181: These classes were made due to the unique serialization to just remove the other elements of the Tally. This simplifies it to just the items to be published. Drastically simplifying the deserialization as well.

tests/integration/test_end_to_end_election.py
```diff
@@ -401,8 +437,68 @@ def publish_results(self) -> None:
             path.exists(RESULTS_DIR),
         )
 
-        if self.REMOVE_OUTPUT:
-            rmtree(RESULTS_DIR)
+    def verify_results(self) -> None:
```
keithrfung, line 440: This test deserializes all the outputs just made into the file directory.

src/electionguard/serializable.py
```diff
@@ -1,21 +1,27 @@
...
-from typing import cast, TypeVar, Generic
 
 T = TypeVar("T")
 
 JSON_FILE_EXTENSION: str = ".json"
 WRITE: str = "w"
+READ: str = "r"
 JSON_PARSE_ERROR = '{"error": "Object could not be parsed due to json issue"}'
+# TODO Issue #??: Jsons library incorrectly dumps class method
+FROM_JSON_FILE = '"from_json_file": {}, '
```
keithrfung, line 24: This is handling a bug with `jsons` that an issue still needs to be made for. I'll update PR accordingly. 

Essentially every output file has this classmethod appended.

src/electionguard/serializable.py
```diff
@@ -30,10 +36,11 @@ def to_json(self, strip_privates: bool = True) -> str:
         :param strip_privates: strip private variables
         :return: the json representation of this object
         """
+        set_serializers()
```
keithrfung, line 39: These were moved into here to reduce turning these on every time they needed to be used.

src/electionguard/ballot.py
```diff
@@ -787,51 +795,155 @@ class BallotBoxState(Enum):
...
+    :param state: ballot box state
     """
 
-    def __post_init__(self,) -> None:
-        super().__post_init__()
+    if len(contests) == 0:
+        log_warning(f"ciphertext ballot with no contests")
 
-        # HACK: ISSUE: #45: Accepted ballots should not have a nonce assoiciated with them
+    # TODO: there's overlap with this code and code in make_ciphertext_ballot: refactor this out
```
AddressXception, line 899: ISSUE #?

src/electionguard/election.py
```diff
@@ -827,46 +861,55 @@ class CiphertextElectionContext(Serializable):
...
+    """
+    Makes a CiphertextElectionContext object.
+
+    :param number_of_guardians: The number of guardians necessary to generate the public key
+    :param quorum: The quorum of guardians necessary to decrypt an election.  Must be less than `number_of_guardians`
+    :param elgamal_public_key: the public key of the election
+    :param description_hash: the hash of the election metadata
+    """
+
+    # What's a crypto_base_hash?
```
keithrfung, line 885: This needs to get fixed.

src/electionguard/group.py
```diff
@@ -125,49 +125,53 @@ def __str__(self) -> str:
...
-def int_to_p_unchecked(i: int) -> ElementModP:
+def int_to_p_unchecked(i: Union[str, int]) -> ElementModP:
     """
     Given a Python integer, returns an ElementModP. Allows
     for the input to be out-of-bounds, and thus creating an invalid
     element (i.e., outside of [0,P)). Useful for tests or if
     you're absolutely, positively, certain the input is in-bounds.
     """
-    return ElementModP(mpz(i))
+    m = mpz(int(i))
```
keithrfung, line 173: Want to create some consistency with these so that all the methods look the same.

src/electionguard/serializable.py
```diff
@@ -48,11 +55,23 @@ def to_json_file(
         """
         write_json_file(self.to_json(strip_privates), file_name, file_path)
 
+    @classmethod
+    def from_json_file(cls, file_name: str, file_path: str = "") -> T:
```
keithrfung, line 59: This could use just the `file_path` to reduce the length of this method. This concatenation could be moved out. I only kept it because the write had that but it could be said that neither need it.

src/electionguard/serializable.py
```diff
@@ -98,5 +112,10 @@ def set_deserializers() -> None:
     set_validator(lambda q: q.is_in_bounds(), ElementModQ)
 
     set_deserializer(
-        lambda usage_string, cls, **_: ProofUsage[usage_string], ProofUsage
+        lambda none, cls, **_: None
```
keithrfung, line 115: This deserialization from "None" is critical. Perhaps we could change it to serialize to `null` as well and have a deserialization from `null` to None but regardless this needs to exist for Ballots to be deserialized.

src/electionguardtest/election.py
```diff
@@ -535,7 +536,7 @@ def election_descriptions(
     end_date = start_date
 
     return ElectionDescription(
-        election_scope_id=draw(languages()),
+        election_scope_id=draw(emails()),
```
keithrfung, line 539: Verify this.

tests/integration/test_end_to_end_election.py
```diff
@@ -90,8 +117,7 @@ def test_end_to_end_election(self) -> None:
         self.step_2_encrypt_votes()
         self.step_3_cast_and_spoil()
         self.step_4_decrypt_tally()
-        self.compare_results()
-        self.publish_results()
+        self.step_5_publish_and_verify()
```
keithrfung, line 120: My expectation is that at some point we will add a 
`5_Publish_and_verify.md` to the repository. It could be argued this PR needs to include that.

src/electionguard/publish.py
```diff
@@ -70,7 +68,7 @@ def publish(
         ballot.to_json_file(ballot_name, BALLOTS_DIR)
 
     make_directory(SPOILED_DIR)
-    for ballot in ciphertext_tally.spoiled_ballots.values():
+    for ballot in spoiled_ballots:
```
AddressXception, line 71:
```
$ make generate-sample-data
pipenv run python src/electionguardtest/sample_generator.py
Traceback (most recent call last):
  File "src/electionguardtest/sample_generator.py", line 118, in <module>
    ElectionSampleDataGenerator().generate()
  File "src/electionguardtest/sample_generator.py", line 101, in generate
    publish(
  File "/Users/mwilhelm/Documents/infernored/microsoft/electionguard-python/src/electionguard/publish.py", line 71, in publish
    for ballot in spoiled_ballots:
TypeError: 'CiphertextTally' object is not iterable
make: *** [generate-sample-data] Error 1
```

PR #140, opened 2020-08-05
✅   Balerion HotFix for Release Build Workflow

### LINKED ISSUE #121: Speed Up CI for Pull Request Workflow
The CI is pretty long for the PR workflow. This could be reduced by only having the Windows and Mac workflows run through the integration tests instead of through the entire test suite since this is already tested on Linux and on the release build.

### LINKED ISSUE #141: Dependency Graph fails on Release Build
This is related to a change in pydeps where the output is displayed by default. This causes the build to get stuck. 

This brings into play a different problem where the release order is not accurate as well. The pypi package should be the last thing completed. Both these issues should be fixed at the same time.

### LINKED ISSUE #142: Update version to v1.1.1
Update version number to 1.1.1 in setup.py

keithrfung:
### Issue

Fixes #121 
Fixes #141 
Fixes #142 

### Description
- Reduce MacOS and Windows to only run integration tests. 
- Add make file to run all integration tests
- Add simpler commands to make file
- Fix dependency graph command issue
- Increment release number

### Testing
- CI should pass. See example.
https://github.com/microsoft/electionguard-python/runs/951389340?check_suite_focus=true

### Checklist

💚Thank you!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,7 +49,7 @@ jobs:
         run: make environment
       - name: Install Module
         run: make install validate
-      - name: Test
+      - name: Full Test & Coverage
         run: make coverage
 
   mac_check:
@@ -68,8 +68,8 @@ jobs:
         run: make environment
       - name: Install Module
         run: make install validate
-      - name: Test
-        run: make coverage
+      - name: Integration Tests
+        run: make test-integration
 
   windows_check:
     name: Windows Check
@@ -89,9 +89,8 @@ jobs:
         run: make environment
       - name: Install Module
         run: make install validate
-      - name: Test
-        continue-on-error: true #TODO Remove Window tests bypass
-        run: make test
+      - name: Integration Tests
+        run: make test-integration
 
   release:
     name: Release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -117,6 +117,13 @@ jobs:
           release_name: Release ${{ env.PACKAGE_VERSION }}
           draft: false
           prerelease: false
+      - name: Dependency Graph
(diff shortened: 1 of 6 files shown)


Pipfile
```diff
@@ -4,6 +4,7 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
+atomicwrites = "*"
```
keithrfung, line 7: This is required for windows tests. `pytest` requires it but it is not installed by default.

Makefile
```diff
@@ -115,8 +119,18 @@ docs-deploy:
...
+	mkdocs gh-deploy --force
+
 dependency-graph:
-	pipenv run pydeps --max-bacon 2 -o dependency-graph.svg src/electionguard
+	pipenv run pydeps --noshow --max-bacon 2 -o dependency-graph.svg src/electionguard
+
+dependency-graph-ci:
+	sudo apt install graphviz
+	pip install pydeps
+	pydeps --noshow --max-bacon 2 -o dependency-graph.svg src/electionguard
```
keithrfung, line 133: The lack of this `--no-show` is what caused the build to fail.

.github/workflows/release.yml
```diff
@@ -118,6 +117,13 @@ jobs:
           release_name: Release ${{ env.PACKAGE_VERSION }}
           draft: false
           prerelease: false
+      - name: Dependency Graph
```
keithrfung, line 120: Moved the order of this up so it wouldn't block the rest of the build after pypi is pushed.

keithrfung:
Please review.

rc-ms:
This looks great!  Thank you for your patience Keith!

Issue #126, opened 2020-07-27
Change JSON Serialization Library


keithrfung:
The existing implementation uses `jsons` library. Changing to the base library `json` instead would simplify and allow some other important usage case like not sorting the output property orders that isn't supported in `jsons`.

- [ ] Remove usage of mpz at low level to simplify other serializing libraries being able to read the classes
- [x] Remove usage of NamedTuples in favor of dataclasses which are typically less problematic with parsers
- [x] Remove any usage of specialized dataclasses where it is not necessary. 
- [ ] Remove publish and methods that soak in all the ballots in favor of just showcasing what classes can be serialized. 
- [ ] Add pydantic example to the utils to showcase how to use an outside sourced serializer

This card should remove the need for other cases such as:
- #96 
- #368

PR #154, opened 2020-08-20
serialize "big" integers as base64 strings while "small" integers stay as integers

### LINKED ISSUE #153: Better handling of JSON serialization of large integers
Right now, the JSON serialization output for bigints sometimes uses strings and sometimes uses ints. This isn't a problem when working in Python3, since all ints are bigints. The rest of the world isn't so pleasant. If we want our structures to play nicely in JavaScript and elsewhere, we should make sure that ElementModP and ElementModQ always serialize to strings.

### LINKED ISSUE #155: JSON serialization of numbers should consider using base64 encoding
In base 10, you get 3.3 bits of information per byte. In base 64, you'd get 6 bits of information per byte. Changing the serializer for elements mod p and q to use base64 strings would thus shrink data files significantly, probably to around 55-58% of the original size. This is much, much easier than trying to rig up gzip or any other sort of general compression, and would require a very modest code change, probably only impacting `set_serializers` and `set_deserializers`.

danwallach:
### Issue

Fixes #153 
Fixes #155

### Description
So far as I can tell, this is the one place where we were writing out bigints as JSON integers instead of JSON strings. The fix is to wrap the constants in `ElementModP`, which already has correct treatment by our serialization process.

### Testing
All tests still pass just fine. Seems like this particular structure isn't much used in ElectionGuard-Python. It's used by Arlo-e2e as kind of version-check. If an election is written out with different constants than the current library is using, that's flagged as an error.

### Checklist

💚Thank you!

### DIFF
--- a/src/electionguard/serializable.py
+++ b/src/electionguard/serializable.py
@@ -112,8 +112,9 @@ def int_to_maybe_base64(i: int) -> Union[str, int]:
 
 def maybe_base64_to_int(i: Union[str, int]) -> int:
     """
-    Given a maybe-encoded big-endian base64-encoded non-negative integer, such as
-    might have been returned by `int_to_base_64_maybe`, returns that integer, decoded.
+    Given a maybe-encoded big-endian base64 non-negative integer, or just a regular
+    integer, such as might have been returned by `int_to_maybe_base64`, returns
+    that integer, decoded.
     :param i: a base64-encode integer, or just an integer
     :return: an integer
     """
--- a/src/electionguard/serializable.py
+++ b/src/electionguard/serializable.py
@@ -95,7 +95,7 @@ def int_to_maybe_base64(i: int) -> Union[str, int]:
     Given a non-negative integer, returns a big-endian base64 encoding of the integer,
     if it's bigger than `ENCODE_THRESHOLD`, otherwise the input integer is returned.
     :param i: any non-negative integer
-    :return: a string in base-64 or just the input integer, if it's "small".
+    :return: a string in base64 or just the input integer, if it's "small".
     """
     assert i >= 0, "int_to_maybe_base64 does not accept negative numbers"
 
@@ -115,7 +115,7 @@ def maybe_base64_to_int(i: Union[str, int]) -> int:
     Given a maybe-encoded big-endian base64 non-negative integer, or just a regular
     integer, such as might have been returned by `int_to_maybe_base64`, returns
     that integer, decoded.
-    :param i: a base64-encode integer, or just an integer
+    :param i: a base64-encoded integer, or just an integer
     :return: an integer
     """
 
--- a/src/electionguard/serializable.py
+++ b/src/electionguard/serializable.py
@@ -1,7 +1,8 @@
+from base64 import b64encode, b64decode
 from dataclasses import dataclass
 from datetime import datetime
 from os import path
-from typing import cast, TypeVar, Generic
+from typing import cast, TypeVar, Generic, Union, Final
 
 from jsons import (
     dumps,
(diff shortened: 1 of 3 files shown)


danwallach:
I'm playing with doing this in general, and it runs into a problem where we're using small integers for fields like `sequence_order`, `number_elected`, and other such things. The base64 encoding of 0 and 1 are, respectively, `AA==` and `AQ==`. This suggests that we don't want to encode *all* integers. Just the big ones.

My current thinking is that I still need the `ElementModP` wrapper within `ElectionConstants`, but that I can make base64 encoding otherwise work in a fairly transparent fashion. Hacking on this now.

danwallach:
This worked out better than I thought it would. I set a threshold, 100 million. Anything below that shows up in the JSON as an integer. Anything about that shows up as a base64-encoded string. All tests passing!

src/electionguard/serializable.py
```diff
@@ -86,14 +87,54 @@ def write_json_file(json_data: str, file_name: str, file_path: str = "") -> None
         json_file.write(json_data)
 
 
+ENCODE_THRESHOLD: Final[int] = 100_000_000
+
+
+def int_to_maybe_base64(i: int) -> Union[str, int]:
```
keithrfung, line 93: This is an integer tool so this should be in the `.utils`

The name is also quite questionable. The word `maybe` should not be used in a method name.

src/electionguard/serializable.py
```diff
@@ -86,14 +87,54 @@ def write_json_file(json_data: str, file_name: str, file_path: str = "") -> None
...
+    """
+    Given a non-negative integer, returns a big-endian base64 encoding of the integer,
+    if it's bigger than `ENCODE_THRESHOLD`, otherwise the input integer is returned.
+    :param i: any non-negative integer
+    :return: a string in base64 or just the input integer, if it's "small".
+    """
+    assert i >= 0, "int_to_maybe_base64 does not accept negative numbers"
```
keithrfung, line 100: We do not want `assert` being hit in production functions outside of tests. This likely should just throw an out of range error. This is the correct behavior.

src/electionguard/serializable.py
danwallach: The use of `maybe` was to convey that small numbers go through while large numbers get encoded. I'm open to other names that convey this sentiment.

src/electionguard/serializable.py
danwallach: Can do.

danwallach:
With respect to the rightness of the approach: the threshold is largely an aesthetic choice. Anything bigger than the cutoff gets encoded.

In terms of where this switches to making dev's lives harder in other languages, those thresholds are when you get greater than MAX_INT in your local representation. So, the threshold we choose should be smaller than any common value of MAX_INT. 

I'm not especially worried about MAX_INT on 16-bit datatypes, since even the 16-bit computers of ancient days gone by, like my old Amiga 1000, were perfectly capable of managing 32-bit integers. In software, but still. As such, the first MAX_INT that I'd say really matters here is 2^31 - 1. That's the top of a C signed integer.

Somewhere, somehow, there's an awful LISP implementation that might borrow one of those 32 bits for type metadata, so we could conservatively say that 2^30 - 1 is a conservative threshold. 2^30 is approximately 1 billion. I set the cutoff lower than that.

So then, why not just convert all integers to base64 strings? Wouldn't that be nice and easier? Consider the following JSON excerpt from an election definition that I'm working with:
```
    {
      "ballot_selections": [
        {
          "candidate_id": "s0000081",
          "object_id": "s0000081",
          "sequence_order": 81
        },
        {
          "candidate_id": "s0000082",
          "object_id": "s0000082",
          "sequence_order": 82
        }
      ],
      "ballot_title": {
        "text": [
          {
            "language": "en",
            "value": "Proposition 112 (Statutory)"
          }
        ]
      },
      "electoral_district_id": "gpunit0000034",
      "name": "Proposition 112 (Statutory)",
      "number_elected": 1,
      "object_id": "contest0000034",
      "sequence_order": 34,
      "vote_variation": "one_of_m"
    },
```

Those `sequence_order` and `number_elected` fields would be unpleasant for a human to read if they were base64 encoded. My current solution keeps all of them nice and legible.

src/electionguard/serializable.py
danwallach: Also, these functions are specifically for use in serialization, and not anywhere else. I don't buy the argument to move them.

keithrfung:
I think this needs to be revisit holistically with #126 . Let's close this and revisit as a whole push.

PR #156, opened 2020-08-31
✨ Open up serialization to reduce clients from depending on jsons

### LINKED ISSUE #157: Reduce external clients dependency on jsons
`electionguard-python` has some specific usage of deserialization and serialization that uses the python package jsons. This binds certain clients to use `jsons` in their implementation. There should be an easier way to do this.

### LINKED ISSUE #158: Iterate to Release 1.1.3
Migrate to next version `1.1.3`

keithrfung:
### Issue

Fixes #158 
Fixes #157 

### Description
This is to allow clients more features using jsons without directly interacting with it until a different solution can be accomplished. This most importantly includes handling exporting to an object. Especially useful in web apis. 

This includes the version bump to facilitate an easy increment. 

### Testing
- `test_serialize`

### DIFF
--- a/src/electionguard/serializable.py
+++ b/src/electionguard/serializable.py
@@ -1,11 +1,13 @@
 from dataclasses import dataclass
 from datetime import datetime
 from os import path
-from typing import cast, TypeVar, Generic
+from typing import Any, cast, TypeVar, Generic
 
 from jsons import (
+    dump,
     dumps,
     NoneType,
+    load,
     loads,
     JsonsError,
     set_deserializer,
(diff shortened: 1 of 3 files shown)


src/electionguard/serializable.py
```diff
@@ -22,7 +24,7 @@
 READ: str = "r"
 JSON_PARSE_ERROR = '{"error": "Object could not be parsed due to json issue"}'
 # TODO Issue #??: Jsons library incorrectly dumps class method
-FROM_JSON_FILE = '"from_json_file": {}, '
+KEYS_TO_REMOVE = ["from_json", "from_json_file", "from_json_object"]
```
keithrfung, line 27: This is an extremely frustrating part of `jsons`. It get confused on class_methods and these keys need to get removed or they show up in the exported object.

PR #163, opened 2020-09-08
Feature/hex_hashes

### LINKED ISSUE #153: Better handling of JSON serialization of large integers

pamtaro:
### Issue
Related to #153 

### Description
These changes will address converting how a bignum (in the form of bytes) will be serialized in to the hash functions (since we're only hashing strings anyway). Hex is easy to convert to and from binary in any language, so instead of using a string representation of the base10 bignum, we just convert the binary to it's hex representation.

### Testing
`make test` to run all tests should still  pass with these changes

### Checklist

💚Thank you!


AddressXception (approved):

src/electionguard/rsa.py
```diff
@@ -95,8 +93,8 @@ def rsa_decrypt(encrypted_message: str, private_key: str) -> Optional[str]:
         plaintext = rsa_private_key.decrypt(ciphertext, PKCS1v15())
     except ValueError:
         return None
-    integer = int.from_bytes(plaintext, BYTE_ORDER)
-    return str(integer)
+    hex_str = plaintext.hex()
```
keithrfung, line 96: Solid improvement.

keithrfung (approved):

PR #172, opened 2020-09-10
Refactor decryption to support individual ballots

### LINKED ISSUE #152: Threshold decryption issues with multiple spoiled ballots
When multiple ballots are spoiled and fewer than the total number of guardians are used to decrypt, tally decryption fails due to an incorrect proof.

This only seems to happen when the following are true:
- More than one ballot is spoiled
- We need to compensate for missing guardians

**TO REPRODUCE:**
[In `test_decryption_mediator.py`](https://github.com/microsoft/electionguard-python/blob/main/tests/test_decryption_mediator.py#L179), spoil the first ballot instead of casting, i.e.:
```python
        ballot_store = BallotStore()
        ballot_box = BallotBox(self.metadata, self.context, ballot_store)
        ballot_box.spoil(encrypted_fake_cast_ballot)
        ballot_box.spoil(encrypted_fake_spoiled_ballot)
```

OR

[Spoil the additional ballots](https://github.com/microsoft/electionguard-python/blob/main/tests/test_decryption_mediator.py#L184):

```python
        # Cast some more fake ballots
        for fake_ballot in self.more_fake_encrypted_ballots:
            ballot_box.spoil(fake_ballot)
```

and then run `test_get_plaintext_tally_compensate_missing_guardian_simple`, which produces warnings about invalid proofs, one guardian being compensated more than once, and ultimately fails to decrypt.

*Note: this was originally found when generating sample data, but is easier to reproduce with the decryption mediator tests.*

### LINKED ISSUE #164: 🔍 Expose decryption methods for individual ballots
The electionguard-web-api is going to be supporting ballot decryption separately from tally decryption, due to the volume of data that would otherwise be passed.

To help with this, we should refactor the `decryption` functions and possibly the `decryption_mediator` to better expose functionality for decrypting individual ballots, while leaving the tally-centric function signatures unchanged.

rkorsak:
### Issues

Fixes #152 
Closes #164 

### Description

✨  Exposes new functions to:
- Compute a decryption share for a ballot
- Compute a compensation share for a ballot
- Reconstruct a missing share for a ballot
- Decrypt a ballot

Examples of using these functions can be found in the decryption mediator tests.

🐛 This also happens to fix a bug in tally share reconstruction whenever multiple ballots were spoiled.

### Testing
Tests have been added for ballot decryption, and updated to include multiple spoiled ballots.

### Checklist

💚Thank you!

### DIFF
--- a/tests/test_decryption_mediator.py
+++ b/tests/test_decryption_mediator.py
@@ -13,11 +13,16 @@ from electionguard.ballot_box import BallotBox, BallotBoxState
 from electionguard.decrypt_with_shares import (
     decrypt_selection_with_decryption_shares,
     decrypt_spoiled_ballots,
+    decrypt_ballot,
 )
 from electionguard.decryption import (
     compute_decryption_share,
+    compute_decryption_share_for_ballot,
     compute_decryption_share_for_selection,
+    compute_compensated_decryption_share_for_ballot,
     compute_compensated_decryption_share_for_selection,
+    compute_lagrange_coefficients_for_guardians,
+    reconstruct_decryption_ballot,
 )
 from electionguard.decryption_mediator import DecryptionMediator
 from electionguard.decryption_share import (
@@ -29,8 +34,8 @@ from electionguard.election import (
     InternalElectionDescription,
 )
 from electionguard.election_builder import ElectionBuilder
-from electionguard.encrypt import EncryptionDevice, EncryptionMediator, encrypt_ballot
 from electionguard.election_polynomial import compute_lagrange_coefficient
+from electionguard.encrypt import EncryptionDevice, EncryptionMediator, encrypt_ballot
 
 from electionguard.group import (
     int_to_q_unchecked,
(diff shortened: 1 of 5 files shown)


tests/test_decryption_mediator.py
```diff
@@ -454,6 +475,95 @@ def test_decrypt_selection_all_present(self):
             self.expected_plaintext_tally[first_selection.object_id], result.tally
         )
 
+    def test_decrypt_ballot_compensate_all_guardians_present(self):
```
rkorsak, line 478: I felt a little weird putting these in the decryption mediator, but there was precedent for tests that directly exercise the decryption functions.

I think it would be ideal in the longer term to pull out some of the shared setup into shared test helpers, so we could move these out.

AddressXception: have you met the [ElectionguardTest](https://github.com/microsoft/electionguard-python/tree/main/src/electionguardtest) module?

AddressXception (approved):
looks good

PR #189, opened 2020-09-28
🕙 Use unix epoch to generate ticks

### LINKED ISSUE #190: 🐛 Datetime tick generation is inconsistent
When a ballot is encrypted using this library, and a `timestamp` is not provided, the tracking hash is computed with `utils.to_ticks(datetime.utcnow())`.  Both `to_ticks` and `utcnow` have some problems:

- `to_ticks` only works on naive timezone-unaware datetimes (e.g. created with `now` or `utcnow`).  Using it on deserialized ISO-8601 timestamps will fail
- `to_ticks` is based on the date `0001-01-01T00:00:00` in naive timezone-unaware form, which:
  - creates very large tick values and risks int overflow
  - is the absolute minimum that datetime can represent, and attempts to normalize to a timezone cause errors
- `to_ticks` returns the number of seconds * 10e7, which is one one hundred millionth of a second - in between microseconds and nanoseconds.  This inflates the size of the tick number massively, and is a non-standard unit.
- `datetime.utcnow` produces the date/time in UTC, [but doesn't capture the timezone](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow).  Most other `datetime` logic (including the time math currently in `to_ticks`) assumes this is a local time, and performs the wrong calculations.

To address this, we've decided to switch to using the unix epoch (seconds since January 1, 1970 UTC).  It's unambiguous and very common.

rkorsak:
Fixes #190 

### Description
- Updates the `utils.to_ticks` function to use the unix epoch (seconds since January 1, 1970 UTC)
- Replaces the use of `datetime.utcnow` with `datetime.now`, as [`utcnow` is considered harmful](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) and ironically produces the wrong epoch timestamp.

Note: sample data will need to be regenerated, but we have several other hashing/format changes coming and should address that in one sweep.

### Testing
Unit tests for `utils.to_ticks` have been augmented to cover more flavors of datetime.

### Checklist

💚 Thank you!

### DIFF
--- a/src/electionguard/utils.py
+++ b/src/electionguard/utils.py
@@ -72,12 +72,12 @@ def to_ticks(date_time: datetime) -> int:
     :return: number of ticks
     """
 
-    epoch_milliseconds = (
+    ticks = (
         date_time.timestamp()
         if date_time.tzinfo
         else date_time.astimezone(timezone.utc).timestamp()
     )
-    return int(epoch_milliseconds)
+    return int(ticks)
 
 
 def space_between_capitals(base: str) -> str:
--- a/src/electionguard/utils.py
+++ b/src/electionguard/utils.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timezone
 from os import mkdir, path
 from re import sub
 from typing import Callable, Optional, TypeVar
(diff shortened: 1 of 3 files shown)


src/electionguard/utils.py
```diff
@@ -66,14 +66,18 @@ def flatmap_optional(optional: Optional[T], mapper: Callable[[T], U]) -> Optiona
 
 def to_ticks(date_time: datetime) -> int:
     """
-    Return the number of ticks for a date time
+    Return the number of ticks for a date time.
+    Ticks are defined here as number of seconds since the unix epoch (00:00:00 UTC on 1 January 1970)
```
keithrfung, line 70: Appreciate the verbage here.

keithrfung (dismissed):
looks good to me. 

Wouldn't mind second review here @AddressXception

src/electionguard/utils.py
```diff
@@ -66,14 +66,18 @@ def flatmap_optional(optional: Optional[T], mapper: Callable[[T], U]) -> Optiona
...
     :param date_time: Date time to convert
     :return: number of ticks
     """
-    t0 = datetime(1, 1, 1)
-    seconds = int((date_time - t0).total_seconds())
-    ticks = seconds * 10 ** 7
-    return ticks
+
+    epoch_milliseconds = (
```
rkorsak, line 75: Rejecting my own PR... these aren't actually milliseconds, but seconds as a float (allowing fractional seconds).  I'll rename.

keithrfung (approved):

PR #199, opened 2020-10-01
Allow Scheduler instance to be injected into decryption functions

### LINKED ISSUE #198: ✨ Allow Scheduler to be injected into decryption functions
Many of the functions in `electionguard.decryption` make use of a new `Scheduler` for parallel processing.  While it's intended to be used as a singleton, we're noticing that right now we are always creating new `Scheduler`s as needed, and it causes problems in our upstream `electionguard-web-api` code - namely, constructing multiple ad-hoc `Scheduler` instances causes our web server to read the child processes in a way that makes it spontaneously shutdown (see https://github.com/microsoft/electionguard-web-api/issues/80).

A complete fix may involve reassessing the `Singleton` helper and change our usage of `Scheduler` to better enforce its singleton nature.  In the meantime, we can get some mileage out of simply adding an `Optional[Scheduler]` argument to all decryption functions which may accept it.  Existing uses of the functions will remain unchanged, but clients can choose to maintain the lifespan of their own `Scheduler` instances as needed.

rkorsak:
### Issue
Closes #198 

### Description
Adds optional arguments to allow clients to manually manage the Scheduler lifecycle.  See notes on the issue itself for more thoughts.

### Testing
Existing tests all pass.

### Checklist

💚 Thank you!

### DIFF
--- a/src/electionguard/decryption.py
+++ b/src/electionguard/decryption.py
@@ -40,7 +40,10 @@ ELECTION_PUBLIC_KEY = ElementModP
 
 
 def compute_decryption_share(
-    guardian: Guardian, tally: CiphertextTally, context: CiphertextElectionContext,
+    guardian: Guardian,
+    tally: CiphertextTally,
+    context: CiphertextElectionContext,
+    scheduler: Optional[Scheduler] = None,
 ) -> Optional[TallyDecryptionShare]:
     """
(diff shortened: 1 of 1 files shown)


keithrfung (approved):
Good quick minimal solution!

PR #213, opened 2020-10-02
Use consistent tracking code format

### LINKED ISSUE #82: Minimize the length of the tracker
The tracker generates words per byte and this can lead to a quite long phrase. It would perhaps be more helpful to abbreviate this into a shorter piece by using another tool other than bytes.

AddressXception:
### Issue
*Link your PR to an issue*

Fixes #82 

### Description
Follow Josh's algorithm for the tracking code

### Testing
`make test`

### Checklist

💚Thank you!

### DIFF
--- a/src/electionguard/tracker.py
+++ b/src/electionguard/tracker.py
@@ -1,5 +1,5 @@
 from typing import List, Optional
-
+from binascii import hexlify
 from .hash import hash_elems
 from .group import ElementModQ, q_to_bytes, bytes_to_q
 from .words import get_word, get_index_from_word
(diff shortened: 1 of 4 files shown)


src/electionguard/tracker.py
```diff
@@ -39,32 +39,29 @@ def tracker_hash_to_words(
...
+        # Select 4 bytes for the segment
+        first = segments[i]
+        second = segments[i + 1]
+        third = segments[i + 2]
+        fourth = segments[i + 3]
 
+        # word is byte(1) + 1/2 of byte(2)
+        word_part = get_word((first * 16 + (second >> 4)))
 
-def tracker_words_to_hash(
```
keithrfung, line 53: The `electionguard-web-api` currently exposes this. Is there no case where a dev may search by the tracker_code but lookup by tracker_hash? Or do we want to add a ticket for someone to create this function for `hacktoberfest`

CC @rkorsak

keithrfung (changes requested):
Lint failing. 

Respond to comments.

src/electionguard/tracker.py
AddressXception: It's not used anywhere in the protocol and specifically it is not used in generating any cryptographic values.  i think it makes sense to discourage using the tracking code anywhere except as a human-readable lookup value in the published results.  further, this representation is entirely arbitrary.  My vote is to remove the endpoint from the web api

keithrfung: Fair point. Just denoting it.

keithrfung (approved):

PR #225, opened 2020-10-16
updated ballot_name -> name, per Google specs (#219)

### LINKED ISSUE #219: Change "ballot_name" in Party class to "name"
Currently the sample manifests contains the property `name` for a `party`, but the `Party` class in Python contains the property called `ballot_name` and is hashing against this property. If you click on the [link to the Google specs for "Party"](https://developers.google.com/elections-data/reference/party),  it also uses `Name` as the reference... so the Python class here is incorrect and needs to change to match the specs and call it the property `name` (like in the example manifests). 

image

Note: this also implies changes that will ripple to the sample generator as well as to the electionguard-web-api's sample data.

markman123:
### Issue
*Link your PR to an issue*

Fixes #219 

### Description
Updates references of `ballot_name` to the updated and correct `name`

### Testing
Determine that all references to `ballot_name` are now `name`

### Checklist

💚Thank you!


keithrfung:
@markman123 Looking good. This just needs `pipenv run black .` to pass the linter.

markman123:
> 
> 
> @markman123 Looking good. This just needs `pipenv run black .` to pass the linter.

I did do that again: 
image

is there a different config?

keithrfung:
@markman123 Sorry on the late response. I'm fairly certain this is an issue with the version of `black` installed. I've had this happen before. There are newer versions where it will autocorrect to a different format. We should lock the version to prevent this issue. I'm running version `19.10b0`. I'm guessing the GitHub Action has grabbed latest. Which version are you running?

keithrfung:
https://github.com/microsoft/electionguard-python/pull/235 Should resolve the build issues on this PR.

src/electionguard/publish.py
```diff
@@ -64,13 +64,13 @@ def publish(
 
     make_directory(BALLOTS_DIR)
     for ballot in ciphertext_ballots:
-        ballot_name = BALLOT_PREFIX + ballot.object_id
```
keithrfung, line 67: These instances do not need to be changed. These are unrelated to `Party`. It just happens to be named `ballot_name` but it is the name of the file.

tests/integration/test_end_to_end_election.py
```diff
@@ -459,17 +459,17 @@ def verify_results(self) -> None:
 
         ciphertext_ballots: List[CiphertextAcceptedBallot] = []
         for ballot in self.ballot_store.all():
-            ballot_name = BALLOT_PREFIX + ballot.object_id
```
keithrfung, line 462: These instances do not need to be changed. These are unrelated to Party. It just happens to be named ballot_name but it is the name of the file.

keithrfung (changes requested):

keithrfung:
@markman123 Can you take a look at requested changes?

keithrfung (approved):

PR #237, opened 2020-10-29
🔧 Open up logging

keithrfung:
### Issue

Fixes #__ Issue incoming

### Description
Add optional parameter to pass in 
- More handlers
- Disable default logging handlers

### Testing
- This can be tested by checking the log output

### Checklist

💚Thank you!


src/electionguard/logs.py
```diff
@@ -17,12 +17,22 @@ class ElectionGuardLog(Singleton):
 
     __logger: logging.Logger
 
-    def __init__(self) -> None:
+    def __init__(
+        self,
+        handlers: List[logging.Handler] = None,
```
rkorsak, line 22: I may already be rusty at my python, but does this need to be wrapped in `Optional` to accept a `None` value?

rkorsak (commented):
This is my fault, but I'm realizing that we can't put this in the `__init__` unless we make deeper changes to our singleton approach.

The library uses the shared `LOG` instance everywhere, and it's initialized in this file with empty arguments.

I think we may just want to expose a method to add a file handler to an existing logger for now.  It wouldn't address the optional file/stream handlers, but it gets us moving at least.

src/electionguard/logs.py
```diff
@@ -21,8 +21,7 @@ def __init__(self) -> None:
         super(ElectionGuardLog, self).__init__()
 
         self.__logger = logging.getLogger("electionguard")
-        self.__logger.addHandler(self._get_file_handler())
```
keithrfung, line 24: Note: The file handler is removed. It can now be added at any point, but is not the default.

mypy.ini
```diff
@@ -6,7 +6,7 @@ warn_return_any = True
 warn_unused_configs = True
 disallow_untyped_defs = True
 ignore_missing_imports = True
-follow_imports = "silent"
+follow_imports = silent
```
keithrfung, line 9: This was throwing a warning on lint. Apparently it does not need quotes.

AddressXception (approved):

PR #255, opened 2020-11-13
Spoiled Ballots should be PlaintextBallots when publishing

### LINKED ISSUE #205: Spoiled Ballots within publish should be Plaintext
ElectionGuard should be publishing (1) the list of **plaintext** spoiled ballots and (2) all encrypted cast or spoiled ballots aka `AcceptedCiphertextBallots`. Currently, the first part is incorrect in the publish.

For this line of code, the spoiled ballots should be `PlaintextBallot`s
https://github.com/microsoft/electionguard-python/blob/main/src/electionguard/publish.py#L41

This affects two tests `end_to_end_integration` and `test_publish`.

On the integration test, this should actually be passing the `PlaintextTally.spoiled_ballots`
https://github.com/microsoft/electionguard-python/blob/main/tests/integration/test_end_to_end_election.py#L428

Required for #96

jpic:
Attempt to close #205


keithrfung:
@jpic  It has to do with this line. When the object is serialized, `jsons` serializer tries to fill in all the properties using a dictionary. `ballot_style` doesn't exist. We have some thoughts on making more DTO style models but the simple solution is to pass a an empty string to the `ballot_style` in your create method.
https://github.com/microsoft/electionguard-python/blob/daad66b8f273a3d90dae73fa2724669c641560ed/src/electionguard/ballot.py#L592

keithrfung:
The results from this test should be placed in a `results` folder when you run the test so you can look at that and see if the JSON is exported in an unexpected way.

jpic:
Thank you @keithrfung, I was looking at the results, so I'm on the right track, the jsons module is pretty nice too, really like your python code !

jpic:
@keithrfung I had a mistake in the further test logic that I could fix locally, but it turns out that create_plaintext_ballot_from_contests is receiving a list of PlaintextTallyContest arguments and creating the PlaintextBallot with that as contests, shouldn't create_plaintext_ballot_from_contests also convert PlaintextTallyContest into PlaintextBallotContest ?

jpic:
This branch is passing, but I'll leave it up to you to choose how to finish create_plaintext_ballot_from_contests, the thing is that currently it sets vote="?" on every selection, I think we want to decipher it but I don't think we have everything we need in the current scope of this function to achieve that unless I'm missing something.

keithrfung:
@jpic I think there is great progress on this but I think you're running into the thing I was afraid of when we started discussing this on the issue. This is gonna require a much deeper dive. I'm going to close this PR, but only because this issue obviously needs a lot more thought. We super appreciate the effort.

PR #269, opened 2021-01-04
Fix publish results_directory argument

### LINKED ISSUE #258: Publish incorrectly computes the directory in project directory instead of passed in directory
In [publish.py](https://github.com/microsoft/electionguard-python/blob/main/src/electionguard/publish.py), all DEVICES_DIR etc are computed at module evalution time which causes DEVICES_DIR in publish to be "results/devices" even when passing a results_directory different from the default: "results".

I've [hacked cwd in our open source multitenant electionguard platform](https://gitlab.com/electisNGO/electeez/-/blob/master/djelectionguard/views.py#L229-254), but fixing results_directory would be nice, I'd like to make a PR if you don't mind.

jpic:
### Issue

Fixes #258

### Description

In publish(), compute directories during runtime rather than import time, to fix the results_directory argument.

### Testing

Try setting the last argument of publish(), results_directory, which is what the tests now do.

💚Thank you!

### DIFF
--- a/src/electionguard/publish.py
+++ b/src/electionguard/publish.py
@@ -10,14 +10,6 @@ from .tally import PlaintextTally, PublishedCiphertextTally
 from .utils import make_directory
 
 RESULTS_DIR = "results"
-DEVICES_DIR = path.join(RESULTS_DIR, "devices")
-COEFFICIENTS_DIR = path.join(RESULTS_DIR, "coefficients")
-BALLOTS_DIR = path.join(RESULTS_DIR, "encrypted_ballots")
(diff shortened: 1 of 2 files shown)


keithrfung:
This is looking really good. Don't worry about Windows Build. #263 will handle that.

keithrfung:
Also I'm gonna change this to a draft PR instead of labeling it as WIP.

jpic:
Thank you @keithrfung, it's ready for review. Didn't find how to convert it back from Draft, sorry about that.

keithrfung (approved):

PR #302, opened 2021-02-04
Change vote from str to int

### LINKED ISSUE #34: `PlaintextBallotSelection` `plaintext` field should be an `int`
Since we added the `extended_data` field to capture text, the `plaintext` field can be replaced with an integer.

As a string, this filed is flexible and supports an extended range of use cases without having to modify the data structure, but at the expense of added parsing complexity.  Since this field really is just a value that is encrypted, it could be an integer and imply to the consumer that this is an encodable value and that the `extended_data` field should be used for any string representations.  One other consideration is that this decreases the human readability of a plaintext ballot expressed in json (unless we further add parsing complexity).

in such an implementation, for `n-of-m` type contests; the value `0` should always resolve to `False` whereas any other value should resolve to `True`

keithrfung:
### Issue

Fixes #34 

### Description
`vote` was changed from type `str` to `int`. This ripples through to several of the tests but ensures that the values can now be added together. This is a long standing change that has needed to happen for quite some time. 

### Testing
`make test` can be used to regression test the changes.


JohnLCaron:
You could modify the PlaintextBallotSelection class documentation:

"This class accepts a `vote` string field which has no constraints
    in the ElectionGuard Data Specification, but is constrained logically
    in the application to resolve to `True` or `False`.  This implies that the
    data specification supports passing any string that can be represented as
    an integer, however only 0 and 1 is supported for now."

AddressXception (changes requested):
I agree we should update the PlaintextBallotSelection description

AddressXception (approved):

PR #305, opened 2021-02-05
🐛  Adjust Spoiled Ballots to be PlaintextTallies

### LINKED ISSUE #205: Spoiled Ballots within publish should be Plaintext

keithrfung:
### Issue
*Link your PR to an issue*

Fixes #205 

### Description
Spoiled Ballots are now PlaintextTally's that can be exported and imported. 
spoiled_ballots were removed from PlaintextTally to avoid circular reference. 
Tests were updated to ensure that the classes don't overlap.

### Testing
Full suite of tests


src/electionguard/decrypt_with_shares.py
```diff
@@ -133,28 +133,19 @@ def decrypt_tally(
     if contests is None:
         return None
 
-    spoiled_ballots = decrypt_spoiled_ballots(
```
keithrfung, line 136: I separated out the `spoiled_ballots` component in `PlaintextTally` since a Spoiled Ballot can be a `PlaintextTally`.

Long term, this means the CiphertextTally also needs to have the `spoiled` component removed as well.

AddressXception (approved):

src/electionguard/ballot.py
```diff
@@ -367,6 +367,19 @@ def __ne__(self, other: Any) -> bool:
         return not self.__eq__(other)
 
 
+@dataclass
+class CiphertextContest(ElectionObjectBase):
+    """
+    Encrypted contest
```
AddressXception, line 373: can we add some more detail here?

src/electionguard/decryption.py
```diff
@@ -506,74 +416,30 @@ def compute_lagrange_coefficients_for_guardians(
...
-    missing_guardians: DataStore[MISSING_GUARDIAN_ID, ElectionPublicKey],
-    compensated_shares: Dict[
-        MISSING_GUARDIAN_ID,
-        Dict[AVAILABLE_GUARDIAN_ID, CompensatedTallyDecryptionShare],
-    ],
-    lagrange_coefficients: Dict[
-        MISSING_GUARDIAN_ID, Dict[AVAILABLE_GUARDIAN_ID, ElementModQ]
-    ],
-) -> Optional[Dict[MISSING_GUARDIAN_ID, TallyDecryptionShare]]:
+def compute_lagrange_coefficients_for_guardian(
```
AddressXception, line 419: i think we should move these lagrange coefficient helper functions to the top or the bottom of the file

src/electionguard/decryption_mediator.py
```diff
@@ -42,81 +48,95 @@ class DecryptionMediator:
...
+                f"announce could not generate tally decryption share for {guardian.object_id}"
             )
             return None
+        self._tally_shares[guardian.object_id] = tally_share
 
-        # Submit the share
-        if self._submit_decryption_share(share):
-            self._available_guardians[guardian.object_id] = guardian
-        else:
+        ballot_shares = compute_decryption_share_for_ballots(
```
AddressXception, line 109: maybe adda comment above this code block so it matches the pattern in the rest of the file?

src/electionguard/decryption_mediator.py
```diff
@@ -42,81 +48,95 @@ class DecryptionMediator:
...
 
-        # This guardian removes itself from the
-        # missing list since it generated a valid share
+        self._mark_available(guardian)
+        if not self._validate_missing_guardian_keys(guardian):
+            return None
+
+        return (tally_share, ballot_shares)
+
+    def _mark_available(self, guardian: Guardian) -> None:
```
AddressXception, line 125: let's move the private function to the bottom of the file

src/electionguard/decryption_mediator.py
```diff
@@ -42,81 +48,95 @@ class DecryptionMediator:
...
             log_warning(
-                f"announce could not submit decryption share for {guardian.object_id}"
+                f"announce could not generate ballot decryption share for {guardian.object_id}"
             )
             return None
+        self._ballot_shares[guardian.object_id] = ballot_shares
 
-        # This guardian removes itself from the
-        # missing list since it generated a valid share
+        self._mark_available(guardian)
```
AddressXception, line 119: also here i think a comment would be helpful to provide context and tell the reader what is going on

src/electionguard/decryption_mediator.py
```diff
@@ -132,189 +152,275 @@ def announce(self, guardian: Guardian) -> Optional[TallyDecryptionShare]:
...
+            log_warning("get plaintext tally failed with share length mismatch")
+            return None
+
+        return decrypt_tally(
+            self._ciphertext_tally,
+            self._tally_shares,
+            self._encryption.crypto_extended_base_hash,
+        )
+
+    def _compute_missing_shares_for_tally(
```
AddressXception, line 206: also suggest adding a summary comment to this method so a developer can get intellisense without having to navigate

tests/property/test_tally.py
```diff
@@ -131,76 +131,64 @@ def test_tally_ballot_invalid_input_fails(
                 from_ciphertext_ballot(encrypted_ballot, BallotBoxState.CAST),
             )
 
-        subject = CiphertextTally("my-tally", metadata, context)
```
AddressXception, line 134: do we want to move away from this pattern permanently?  I generally like to use generic names for local vars inside tests, things like `subject` and `data` and `result` or `expected` and `actual` - things like that - but i could be convinced its better to use descriptive names.  What I like about identifying the `subject` specifically is that no matter where you are in a code block the object under test is always clear.

AddressXception (changes requested):
This is a massive structural improvement over what was here before. I know this one wasn't easy.  I left a few comments that are extremely nit-picky and may not all be relevant.  Set to request changes just to make sure we talk about it first before merge.

nice job on this.

src/electionguard/ballot.py
keithrfung: Yeah. It looks like it got cut off.

keithrfung: If you look at CiphertextSelection, this lines up 1 to 1.

tests/property/test_tally.py
keithrfung: I think when tests are centered a bit better it's useful but I think there is higher usability in describing what we are doing.

keithrfung:
@AddressXception Changes complete.

AddressXception (approved):

PR #313, opened 2021-02-09
Hash Guardian Commitments

### LINKED ISSUE #279: Possible error in generating extended_crpyto_hash
In writing a validator for the specification "3. Election Public-Key Validation":

````
(A) Q̅ = H(Q, K 1,0 , K 1,1 , K 1,2 , ... , K 1,k−1 , K 2,0 , K 2,1 , K 2,2 , ... , K 2,k−1 , ... , K n,0 , K n,1 , K n,2 , ... , K n,k−1 )
(B) K = ∏ ni=1 K i mod p.
````

Which I understand to mean that the extended_crpyto_hash Q̅ has "all of the public guardian commitments 𝐾𝑖,𝑗  hashed together"

However in election.py lines starting at 899:

    crypto_base_hash = hash_elems(
        P, Q, G, number_of_guardians, quorum, description_hash
    )
    crypto_extended_base_hash = hash_elems(crypto_base_hash, elgamal_public_key)

Which I think means that only the Guardian's public keys are used, ie:

````
Q̅ = H(Q, K 1,0 , K 2,0 ... , K n,0 )
Q̅ = H(Q, K)
````

### PR DESCRIPTION
**Pull request by AddressXception**

### Issue

Fixes #279 

### Description

Add Guardian coefficient commitments to extended base hash

generate a commitment hash for guardians when generating the joint public key and propagate the commitment hash as part of a named tuple with the joint public key.  Use the commitment hash as part of the election builder and the cipertext election context, caching the result as part of the ciphertext election context.

### Testing

1. `make test`

### CODE REVIEW

File: src/electionguard/ballot.py [hunk 1]
```diff
@@ -414,6 +414,9 @@ class CiphertextBallotContest(ElectionObjectBase, CryptoHashCheckable):
     ballot_selections: List[CiphertextBallotSelection]
     """Collection of ballot selections"""
 
+    ciphertext: ElGamalCiphertext
```
keithrfung, line 417: We use `ciphertext` for a lot for properties but I don't this is helpful for determining what this is because we use `ciphertext` in so many sections. `accumulated_votes` or `total` would make sense. idk if always need to indicate it's `encrypted` or `ciphertext`

src/electionguard/ballot.py
```diff
@@ -596,6 +608,7 @@ def make_ciphertext_ballot_contest(
         description_hash=description_hash,
         ballot_selections=ballot_selections,
         nonce=nonce,
+        ciphertext=elgamal_accumulation,
```
keithrfung, line 611: This highlights what I'm saying. I feel like this should be a much more meaningful line.

keithrfung (changes requested):

keithrfung (approved):

PR #315, opened 2021-02-12
update election.py reference

### LINKED ISSUE #314: election.py references nonexistent electionguard/wiki
election.py refers several times to

  [ElectionGuard Spec](https://github.com/microsoft/electionguard/wiki)

which does not exist, but gets redirected to 

  https://github.com/microsoft/electionguard

rc-ms:
Fixes #314 

### Description
updating wiki reference to point to top-level spec url in electionguard.vote
https://www.electionguard.vote/spec/Overview/

### Testing
documentation change; no functionality affected


keithrfung (changes requested):
The links should be using https://microsoft.github.io/electionguard NOT https://electionguard.vote

This way if the redirects break this will still work. Also maybe these references shouldn't be here at all?

keithrfung (approved):

keithrfung:
I think the best play for this is to simply remove these references since we can have these outside the code base.

keithrfung:
This will be done here instead.
https://github.com/microsoft/electionguard-python/issues/352

PR #331, opened 2021-03-02
✨  Implement Compact Ballot

### LINKED ISSUE #330: Create a compact form of the ballot
## Feature Request

**Description**
The ballots are by design very bulky.  The contests contain a lot of information. This is an effort to shrink these down for convenience and certain API use cases. The naming scheme we think would be appropriate is a `Compact` ballot. Here are some examples:

`PlaintextBallot` compresses to `CompactPlaintextBallot`
`CompactPlaintextBallot` expands to `PlaintextBallot`

keithrfung:
### Issue
*Link your PR to an issue*

Fixes #330 

### Description
- Add Compact Plaintext and Ciphertext Ballot
- Update Encryption Device and hashing mechanisms
- Isolate `create_ballot_hash` to get a single ballot hash
- Isolate `encrypt_ballot_contests` for using to encrypt all the ballot contests

### Testing
In Progress

### DIFF
--- a/src/electionguard/ballot_compact.py
+++ b/src/electionguard/ballot_compact.py
@@ -162,7 +162,7 @@ def _get_plaintext_contests(
         ):
             selections.append(
                 PlaintextBallotSelection(
-                    selection.candidate_id,
+                    selection.object_id,
                     YES_VOTE if compact_ballot.selections[index] else NO_VOTE,
                     not contest_in_style,
                     compact_ballot.extended_data.get(index),
--- a/src/electionguard/ballot_compact.py
+++ b/src/electionguard/ballot_compact.py
@@ -1,0 +1,181 @@
+from dataclasses import dataclass
+from typing import Dict, List
+
+
+from .ballot import (
+    SubmittedBallot,
+    ExtendedData,
+    PlaintextBallot,
+    PlaintextBallotContest,
+    PlaintextBallotSelection,
+    create_ballot_hash,
+)
+from .ballot_box import BallotBoxState
+from .election import CiphertextElectionContext
+from .encrypt import encrypt_ballot_contests
+from .group import ElementModQ
+from .manifest import (
+    ContestDescriptionWithPlaceholders,
+    InternalManifest,
+)
+from .utils import get_optional
+
+YES_VOTE = 1
+NO_VOTE = 0
+
+
+@dataclass
+class CompactPlaintextBallot:
+    """A compact plaintext representation of ballot minimized for data size"""
+
+    object_id: str
+    style_id: str
+    selections: List[bool]
+    extended_data: Dict[int, ExtendedData]
+
+
+@dataclass
+class CompactSubmittedBallot:
+    """A compact submitted ballot minimized for data size"""
+
+    compact_plaintext_ballot: CompactPlaintextBallot
+    timestamp: int
+    ballot_nonce: ElementModQ
+    previous_code: ElementModQ
+    code: ElementModQ
+    ballot_box_state: BallotBoxState
+
+
+def compress_plaintext_ballot(ballot: PlaintextBallot) -> CompactPlaintextBallot:
+    """Compress a plaintext ballot into a compact plaintext ballot"""
+    selections = _get_compact_selections(ballot)
+    extended_data = _get_compact_extended_data(ballot)
+
+    return CompactPlaintextBallot(
+        ballot.object_id, ballot.style_id, selections, extended_data
+    )
+
+
+def compress_submitted_ballot(
+    ballot: SubmittedBallot,
+    plaintext_ballot: PlaintextBallot,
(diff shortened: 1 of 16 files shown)


src/electionguard/ballot_compact.py
```diff
@@ -0,0 +1,138 @@
...
+    compact_ballot: CompactPlaintextBallot, description: InternalElectionDescription
+) -> List[PlaintextBallotContest]:
+    """Get ballot contests from compact plaintext ballot"""
+    index = 0
+    ballot_style_contests = _get_ballot_style_contests(
+        compact_ballot.style_id, description
+    )
+
+    contests: List[PlaintextBallotContest] = []
+    for description_contest in sorted(
```
keithrfung, line 106: This sort I feel like we don't often do, but I'm ensuring that it is used. Perhaps this should be a utility function.

src/electionguard/ballot_compact.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+@dataclass
+class CompactPlaintextBallot:
+    """A compact plaintext representation of ballot minimized for data size"""
+
+    object_id: str
+    style_id: str
+    selections: List[bool]
+    extended_data: Dict[int, ExtendedData]
```
AddressXception, line 34: I used the string id here but i suppose the index position is better...

src/electionguard/ballot_compact.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+@dataclass
+class CompactSubmittedBallot:
+    """A compact submitted ballot minimized for data size"""
+
+    compact_plaintext_ballot: CompactPlaintextBallot
+    timestamp: int
+    ballot_nonce: int
```
AddressXception, line 43: I'm holding onto the actual `ElementModQ` references in the c++ objects because the serialization needs to export in hex string and its already supported in those objects.

src/electionguard/ballot_compact.py
```diff
@@ -0,0 +1,186 @@
...
+
+    contests: List[PlaintextBallotContest] = []
+    for description_contest in sorted(
+        description.contests, key=lambda c: c.sequence_order
+    ):
+        contest_in_style = (
+            ballot_style_contests.get(description_contest.object_id) is not None
+        )
+
+        # Iterate through selections. If contest not in style, mark placeholder
```
AddressXception, line 161: i broke this out into a separate function called `emplace_missing_values`

src/electionguard/encrypt.py
```diff
@@ -35,18 +35,39 @@ class EncryptionDevice(Serializable):
...
+    """Arbitary string to designate the location of device"""
 
-    def __init__(self, location: str) -> None:
+    def __init__(
+        self,
+        session_id: str,
+        launch_code: int,
+        location: str,
+    ) -> None:
         self.uuid = generate_device_uuid()
```
AddressXception, line 55: im passing this as a constructor parameter... im not sure which way is the better way to do it but we should be consistent.

src/electionguard/encrypt.py
```diff
@@ -35,18 +35,39 @@ class EncryptionDevice(Serializable):
...
         self.location = location
+        self.launch_code = launch_code
```
AddressXception, line 57: ctor order match the hash order

AddressXception (changes requested):

src/electionguard/encrypt.py
keithrfung: if it's a construction parameter it will have to be the last parameter. It's labeled as the first object in the hash.

src/electionguard/encrypt.py
keithrfung: I'm down with this idea but counter intuitive with the notes on uuid.

src/electionguard/ballot_compact.py
keithrfung: index makes more sense since the reference here has to stand alone. This ties selections and extended_data within this compact ballot together.

tests/test_decryption_mediator.py
```diff
@@ -106,7 +111,9 @@ def setUp(self):
         builder.set_commitment_hash(self.joint_public_key.commitment_hash)
         self.internal_manifest, self.context = get_optional(builder.build())
 
-        self.encryption_device = EncryptionDevice("location")
+        self.encryption_device = EncryptionDevice(
+            generate_device_uuid(), "Session", 12345, "Location"
```
keithrfung, line 115: Maybe this should be in the mock section since it is repeated so much?

ElectionFactory.get_encryption_device() ?

AddressXception: i was thinking the same thing actually

AddressXception (approved):

AddressXception (approved):

PR #332, opened 2021-03-04
Refactor Tracker to Ballot Code and Remove Words

### LINKED ISSUE #322: Refactor Tracker
## Issue

**Description**
- [x] Refactor `tracker_hash` to `code` in code
- [x] Refactor tracker has to Verification Code in docs
- [x] Remove `words.py` and associated tests and methods

`tracker_hash` should be refactored to `code` and similar methods should be adjusted to no longer refer it by this name. The use of this has led to confusion with many vendors. The plan is to switch to `ballot.code` which makes a bit more sense in plain english since this ends up being a top level term. 

The tracking words will be removed around the same time. This implementation is a carryover and provides some convenience but the string representation of the code can be adjust to create many useful presentations besides the one currently offered. The words also present a maintenance problem in the long term.

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #322 
Mentioned in Discussion #319 

### Description
 Refactor tracker_hash to code in code
 Refactor tracker has to Verification Code in docs
 Remove words.py and associated tests and methods

### Testing
- Existing / Refactored tests


src/electionguard/ballot.py
```diff
@@ -666,14 +666,14 @@ class CiphertextBallot(ElectionObjectBase, CryptoHashCheckable):
     description_hash: ElementModQ
     """Hash of the election metadata"""
 
-    previous_tracking_hash: ElementModQ
-    """Previous tracking hash or seed hash"""
+    previous_code: ElementModQ
```
AddressXception, line 669: i think we should qualify this as `previous_ballot_code`

src/electionguard/ballot.py
```diff
@@ -858,11 +849,11 @@ def make_ciphertext_ballot(
     object_id: str,
     ballot_style: str,
     description_hash: ElementModQ,
-    previous_tracking_hash: Optional[ElementModQ],
+    previous_code: Optional[ElementModQ],
     contests: List[CiphertextBallotContest],
     nonce: Optional[ElementModQ] = None,
     timestamp: Optional[int] = None,
-    tracking_hash: Optional[ElementModQ] = None,
+    code: Optional[ElementModQ] = None,
```
AddressXception, line 856: same, `ballot_code`.  i think we need to provide at least some context in the parameter name

AddressXception (changes requested):
did the words file actually go somewhere?  I think we should keep it in the tests so we can show how the "friendly name" can be generated but maybe @rc-ms disagrees?

src/electionguard/ballot.py
keithrfung: Ok. I'll use this for the methods. I was favoring brevity.

keithrfung:
@AddressXception We don't need the `words.py` the move for this based on the discussion was removing it from the codebase and providing another simple example at the top **api** level. We have the git history if we need it. This isn't going to exist in the base modules.

src/electionguard/ballot.py
keithrfung: this was changed at the correct method level.

PR #336, opened 2021-03-04
Fix guardian = 0 in test_end_to_end_election

### LINKED ISSUE #335: test_end_to_end_election is using guardian sequence_order = 0.
Guardian sequence_order must be > 0.

JohnLCaron:
Fixes #335

Guardian sequence_order must be > 0.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/electionguard-python/336)

### DIFF
--- a/tests/integration/test_end_to_end_election.py
+++ b/tests/integration/test_end_to_end_election.py
@@ -167,7 +167,7 @@ class TestEndToEndElection(TestCase):
         # Setup Guardians
         for i in range(self.NUMBER_OF_GUARDIANS):
             self.guardians.append(
-                Guardian("guardian_" + str(i), i, self.NUMBER_OF_GUARDIANS, self.QUORUM)
+                Guardian("guardian_" + str(i), i + 1, self.NUMBER_OF_GUARDIANS, self.QUORUM)
             )
 
         # Setup Mediator


keithrfung (changes requested):
One line change so easy to review. 
1. Linter needs to pass. `make auto-lint`
2. Do you think this should be sitting at the guardian level itself? Maybe just check at that level and never allow a `0`? It's an easy check in python with a setter.

JohnLCaron:
I agree that there should be a check in the Guardian to not allow sequence <= 0. If you want I can add here. But not sure what is the cleanest syntax for that in python.

keithrfung:
@AddressXception I think this should be a `ValueError` on the Guardian constructor. What do you think of this?

https://github.com/microsoft/electionguard-python/blob/22663821cc83e1d3fdaaec3079b8c9dcff0d0b1e/src/electionguard/guardian.py#L104

```py
MIN_SEQUENCE_ORDER  =  1
SEQUENCE_ORDER_ERROR = f"Sequence Order must be at least {MIN_SEQUENCE_ORDER}"

if sequence_order <  MIN_SEQUENCE_ORDER:
     throw ValueError(SEQUENCE_ORDER_ERROR)
```

keithrfung:
@JohnLCaron Can you revisit this?

JohnLCaron:
Yes, sorry, my python installation got mangled and I havent made it back to fixing it.

keithrfung:
@JohnLCaron I actually rolled this fix into the Key Ceremony fix since I created a helper method to ensure this doesn't crop up again.

keithrfung:
I'll close the issue once the integration branch is merged in.

JohnLCaron:
thanks, keith.

PR #349, opened 2021-03-29
Key Ceremony Update

### LINKED ISSUE #346: Key Ceremony Rework
## Issue

**Description**
The Key Ceremony has a couple issues that need to be resolved. This issue acts as a catch all for known issues
1. The `sequence_order` and `owner_id` should be present in the keys that are shared.
2. The `Guardian` itself contains too much state and should not be serializable. A `Guardian` should be able to be reconstructed from it's information. 
3. The Guardian should serialize or publish out to a different object. The naming convention currently is `GuardianRecord`
4. The `DecryptionMediator` contains a full `Guardian`. This should not be the case. A `Mediator` should only know about public or shared information.
5. The commitments that are shared in Round 2 with the backups should actually be shared in Round 1. 
6. There should be a purely functional example of the key ceremony in a test to demonstrate that this can be done without the use of a `KeyCeremonyMediator` even if a lower case mediator is still used.

keithrfung:

**Note:**
- This is being merged into an **integration branch** due to its size
- ⚠️ Code coverage is currently 88%. This is going to get resolved. The code cleanup in certain areas reduced some code lines leading to lower coverage. 

## Description
This is a significant refactor to the Key Ceremony and pieces within it. 

### Key Ceremony itself
- Migrate coefficient to `ElectionPublicKey` instead of `ElectionPartialKeyBackup`
- Pass in public key or keypair instead of entire stateful `Guardian`
- Add `share` methods on keys to quickly share the public key of a key pair

### Stateless Decisions
- Ensure a KeyPair of any type has a guardian's id and sequence order 
- Remove the `Guardian`'s in the `KeyCeremonyMediator`

### BoyScouting
- Remove of `Optional` in DataStore
- Use of rename for types particularly for `ElementModP` and `ElementModQ` like `ELECTION_PUBLIC_KEY` types to reduce down references to `group.py`
- Implementation of `KeyCeremonyHelper` to reduce to duplicate code to recreate a `KeyCeremony` for many later tests like decryption

## Code Direction
- Focus on key ceremony being a series of three (possible four) distinguishable rounds
- Reduce of usages of `DataStore` to reduce the dependency on a store that could be replaced with a dictionary
- Reduce references when possible
- Start migrating to a `electionguard-mock` (`electionguardtest`) library
- Start migrating to a stateless electionguard package

## Overall Remaining Work
_work that exists beyond this PR_
- Isolate stateful components
- Isolate mocks
- `BallotBox` needs a stateless implementation
- `DecryptionMediator` needs a stateless implementation

### Issue

Fixes #346 

### Testing

Existing test suite should be sufficient to display most tests. 

`test_functional_key_ceremony` was added to demonstrate an example of the functional key ceremony.

`test_decryption_mediator.py` was broken up to two different tests since it included testing of both `decryption.py` and `decryption_mediator.py`


src/electionguard/decryption_mediator.py
```diff
@@ -84,34 +84,38 @@ def announce(
...
-                self._tally_shares[guardian.object_id],
-                self._ballot_shares[guardian.object_id],
+                self._tally_shares[guardian.id],
+                self._ballot_shares[guardian.id],
             )
 
         # Compute the tally and ballot decryption shares
         tally_share = compute_decryption_share(
-            guardian, self._ciphertext_tally, self._encryption
+            guardian._election_keys,  # pylint: disable=protected-access
```
keithrfung, line 96: Private keys are referenced here in the tests but this is not an optimal case. Here we are using the stateful methods to quickly construct guardians but this method is a meant to be used in a functional capacity.

src/electionguard/key_ceremony.py
```diff
@@ -313,35 +286,22 @@ def verify_election_partial_key_challenge(
     )
 
 
-def hash_coefficient_commitments(
-    commitments: List[ElementModP],
-) -> Optional[ElementModQ]:
-    """
-    Hashes the commitments together
-    :param commitments: the flattened list of commitments of all guardians in sequence order
-    :return: H(K 1,0 , K 2,0 ... , K n,0 )
```
keithrfung, line 322: This needs to be kept. Will be moved.

src/electionguard/election_polynomial.py
```diff
@@ -107,21 +109,32 @@ def compute_lagrange_coefficient(coordinate: int, *degrees: int) -> ElementModQ:
...
+    for (i, commitment) in enumerate(commitments):
         exponent = pow_p(int_to_p_unchecked(exponent_modifier), int_to_p_unchecked(i))
         factor = pow_p(commitment, exponent)
         commitment_output = mult_p(commitment_output, factor)
 
     value_output = g_pow_p(coordinate)
     return value_output == commitment_output
+
+
+def _check_exponent_modifier(exponent_modifier: int) -> None:
```
AddressXception, line 135: why not just call `ElementModQ.is_in_bounds()` instead of making this method?

src/electionguard/guardian.py
```diff
@@ -41,16 +42,41 @@
 from .types import GUARDIAN_ID
 from .utils import get_optional
 
+
+@dataclass
+class GuardianRecord(Serializable):
+    """Published record of Guardian"""
```
AddressXception, line 48: as i mentioned on the call, would be good to have more of a description here so that i can know the context of a guardian record without having to go look it up in the docs

src/electionguard/guardian.py
```diff
@@ -41,16 +42,41 @@
...
+
+    guardian_id: GUARDIAN_ID
```
AddressXception, line 50: similarly we should annotate all of the properties

src/electionguard/guardian.py
```diff
@@ -41,16 +42,41 @@
...
+    sequence_order: int
+    election_public_key: ELECTION_PUBLIC_KEY
+    election_commitments: List[PUBLIC_COMMITMENT]
+    election_proofs: List[SchnorrProof]
+
+
+def publish_guardian_record(election_public_key: ElectionPublicKey) -> GuardianRecord:
```
AddressXception, line 57: same a summary comment would be helpful for context

src/electionguard/guardian.py
```diff
@@ -412,7 +420,7 @@ def all_election_partial_key_backups_verified(self) -> bool:
         return True
 
     # Joint Key
-    def publish_joint_key(self) -> Optional[ElementModP]:
+    def publish_joint_key(self) -> Optional[ELECTION_JOINT_PUBLIC_KEY]:
```
AddressXception, line 423: im of mixed feelings for renaming all of the types from `ElementModP` to descriptive types like `ELECTION_JOINT_PUBLIC_KEY`.  on one hand it improves readability in this code base, on the other hand it is going to take a fair amount of work to make it consistent across all languages.  in any case, there are a number of functions in this file that still return `Tuple[ElementModP,...]` or just `ElementModP` so i'm game for either but i think we should try to fix it in whole files whenever we make the change once in a file.  thoughts?

src/electionguard/key_ceremony.py
```diff
@@ -19,21 +18,53 @@
...
+VERIFIER_ID = str
+
+
+class ElectionPublicKey(NamedTuple):
+    """A tuple of election public key and owner information"""
+
+    owner_id: GUARDIAN_ID
+    sequence_order: int
+    key: ELECTION_PUBLIC_KEY
+    coefficient_commitments: List[PUBLIC_COMMITMENT]
```
AddressXception, line 39: lets pop some comments on these props, specifically to call out that the `key` is a convenience accessor for `coefficient_commitments[0]` so its clear what is happening and also maybe state that the `coefficient_proofs` are also index based and map to the commitments?

src/electionguard/key_ceremony.py
```diff
@@ -19,21 +18,53 @@
...
+    coefficient_proofs: List[SchnorrProof]
+
+
+class ElectionKeyPair(NamedTuple):
+    """A tuple of election key pair, proof and polynomial"""
+
+    owner_id: GUARDIAN_ID
+    sequence_order: int
+    key_pair: ElGamalKeyPair
+    polynomial: ElectionPolynomial
```
AddressXception, line 49: is the polynomial the public or secret values? i forget

src/electionguard/key_ceremony.py
```diff
@@ -313,35 +286,22 @@ def verify_election_partial_key_challenge(
...
 ) -> ElectionJointKey:
     """
     Creates a joint election key from the public keys of all guardians
-    :param election_commitments: a dictionary of commitments keyed by guardian object_id
-    :param election_public_keys: a data store of public keys keyed by guardian object_id
+    :param election_public_keys: all public keys of the guardians
     :return: ElectionJointKey for election
     """
-    public_keys = map(lambda public_key: public_key.key, election_public_keys.values())
+    public_keys = [set.key for set in election_public_keys]
```
AddressXception, line 297: 👍

src/electionguardtest/key_ceremony_helper.py
```diff
@@ -0,0 +1,106 @@
+from typing import List
+
+from electionguard.guardian import Guardian
+from electionguard.key_ceremony import CeremonyDetails, ElectionPartialKeyVerification
+from electionguard.key_ceremony_mediator import GuardianPair, KeyCeremonyMediator
+
+from .identity_encrypt import identity_auxiliary_decrypt, identity_auxiliary_encrypt
+
+
+class KeyCeremonyHelper:
```
AddressXception, line 10: i think this is considerably more readable than what we had before

AddressXception: im also of two minds on files like this, which could clearly be useful in an optional package/module for library consumers and then the actual "test" generators pieces, which are only useful for the tests.  In c++ i create a separate library called 'mocks' that includes all the test generators, but im not quire sure what to do with files like this long term.  perhaps the answer is that we intend to refactor into component modules at some point and we handle this then.

tests/integration/test_end_to_end_election.py
```diff
@@ -166,71 +168,103 @@ def step_1_key_ceremony(self) -> None:
         # Setup Guardians
         for i in range(self.NUMBER_OF_GUARDIANS):
             self.guardians.append(
-                Guardian("guardian_" + str(i), i, self.NUMBER_OF_GUARDIANS, self.QUORUM)
+                Guardian(
```
AddressXception, line 171: i know this was here before but it seems error prone and we've hit this before a few times when the indexing is off.  maybe we should move it into the key ceremony helper and reference any iterators that consume this pattern from there?  or is this part of #336 ?

tests/integration/test_functional_key_ceremony.py
```diff
@@ -0,0 +1,234 @@
...
+    generate_rsa_auxiliary_key_pair,
+    verify_election_partial_key_backup,
+    verify_election_partial_key_challenge,
+)
+from electionguard.types import GUARDIAN_ID
+from electionguardtest.identity_encrypt import identity_auxiliary_encrypt
+
+
+class TestKeyCeremony(TestCase):
+    """Functional tests of key ceremony"""
```
AddressXception, line 23: some more details about why this exists would be helpful

tests/property/test_decryption_mediator.py
```diff
@@ -0,0 +1,327 @@
...
+            ballot_box.cast(fake_ballot)
+        # Spoil some more fake ballots
+        for fake_ballot in self.more_fake_encrypted_spoiled_ballots:
+            ballot_box.spoil(fake_ballot)
+
+        # generate encrypted tally
+        self.ciphertext_tally = tally_ballots(
+            ballot_store, self.internal_manifest, self.context
+        )
+        self.ciphertext_ballots = get_ballots(ballot_store, BallotBoxState.SPOILED)
```
AddressXception, line 181: lol @ the length of this method :-p honestly though, i dont think you should change it because this is a unit test and we are trying to show all of the steps of a linear process.  breaking this out into small functions will decrease readability in my opinion

src/electionguard/election_polynomial.py
keithrfung: This can be done for the high bound but the actual importance of this method is the lower end. There is an active PR that also addresses this. This is in relation to the fact that sequence order cannot be zero.

tests/property/test_decryption_mediator.py
keithrfung: I would like to venture a suggestion that we remove this a different way. This should be handled in `Helper`, `Factory` or `Builder` in the `electionguardtest`/`electionguard-mock` setup. 

`KeyCeremonyHelper.perform_full_ceremony`
`BallotBoxHelper.mock_voting`
`TallyHelper.tally_ballots`

tests/integration/test_end_to_end_election.py
keithrfung: Actually this is a mistake. This should use the KeyCeremonyHelper to create the guardians. That'll reduce the need for #336.

src/electionguardtest/key_ceremony_helper.py
keithrfung: you and me both. I really think this one is slowly building out to a package. 

Purpose 1: Mock the steps in an election for easy 
Purpose 2: Assist in generation and complicated flow for testing library

tests/property/test_decryption_mediator.py
AddressXception: agreed

src/electionguard/guardian.py
keithrfung: I'm not sure how to tackle this. Perhaps this should be a larger issue. I truly think we need to shift to a recognizable type but I don't think every language should necessarily do this. We should discuss this in detail.

src/electionguard/key_ceremony.py
keithrfung: the polynomial is the secret values

src/electionguard/guardian.py
```diff
@@ -99,8 +155,7 @@ def __init__(
         :param nonce_seed: an optional `ElementModQ` value that can be used to generate the `ElectionKeyPair`.
                            It is recommended to only use this field for testing.
         """
```
keithrfung, line 157: As a note, perhaps Guardian would be better if it took in the election keys and auxiliary keys to allow easier reconstruction.

AddressXception (approved):

PR #379, opened 2021-06-22
hash now correctly handles integers, other iterables

### LINKED ISSUE #378: hashing doesn't correctly handle some corner cases
## Issue

**Description**
In experimenting with my "educational" version of ElectionGuard, where I accumulated the whole formula of every operation on ElementModP and Q, I discovered a few misfeatures in the hash function.

- If the value being hashed is the integer 0, that's treated by Python as "false" and is hashed as "null"
- There were some cases where lists weren't being hashed recursively, but were instead being converted to a string and then hashed as a string

None of these impact correctness, since the hash still properly depends on the value in question, but they do impact portability, in the sense that another programming language might properly implement the same hash function with identical behavior. 

The patch is pretty simple, only a few changes in `hash.py`

danwallach:

### Issue
Fixes #378 

### Description
See #378 

### Testing
Minor changes, all tests still pass.

### DIFF
--- a/src/electionguard/hash.py
+++ b/src/electionguard/hash.py
@@ -1,7 +1,14 @@
 from abc import abstractmethod
 from collections.abc import Sequence
 from hashlib import sha256
-from typing import Union, Protocol, runtime_checkable, Sequence as TypedSequence
+from typing import (
+    Iterable,
+    List,
+    Union,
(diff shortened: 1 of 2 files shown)


keithrfung (approved):
lgtm.

PR #387, opened 2021-07-16
🧮 Add new Primes and utilize test primes in testing suite

### LINKED ISSUE #76: Add crypto "test parameters" to speed up unit tests

### LINKED ISSUE #329: ✨ Update Primes
## Issue

This was request by @benaloh and is an update to the primes. The following are the values that should change. 

# p
The large prime p should be the following.

**Decimal:**
```
1044388881413152506691752710716624382579964249047383780384233483283953907971553643537729993126875883902173634017777416360502926082946377942955704498542097614841825246773580689398386320439747911160897731551074903967243883427132918813748016269754522343505285898816777211761912392772914485521155521641049273446207578961939840619466145806859275053476560973295158703823395710210329314709715239251736552384080845836048778667318931418338422443891025911884723433084701207771901944593286624979917391350564662632723703007964229849154756196890615252286533089643184902706926081744149289517418249153634178342075381874131646013444796894582106870531535803666254579602632453103741452569793905551901541856173251385047414840392753585581909950158046256810542678368121278509960520957624737942914600310646609792665012858397381435755902851312071248102599442308951327039250818892493767423329663783709190716162023529669217300939783171415808233146823000766917789286154006042281423733706462905243774854543127239500245873582012663666430583862778167369547603016344242729592244544608279405999759391099769165589722584216017468464576217318557948461765770700913220460557598574717173408252913596242281190298966500668625620138188265530628036538314433100326660047110143
```

**Hex:**
```
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF93C467E37DB0C7A4D1BE3F810152CB56A1CECC3AF65CC0190C03DF34709AFFBD8E4B59FA03A9F0EED0649CCB621057D11056AE9132135A08E43B4673D74BAFEA58DEB878CC86D733DBE7BF38154B36CF8A96D1567899AAAE0C09D4C8B6B7B86FD2A1EA1DE62FF8643EC7C271827977225E6AC2F0BD61C746961542A3CE3BEA5DB54FE70E63E6D09F8FC28658E80567A47CFDE60EE741E5D85A7BD46931CED8220365594964B839896FCAABCCC9B31959C083F22AD3EE591C32FAB2C7448F2A057DB2DB49EE52E0182741E53865F004CC8E704B7C5C40BF304C4D8C4F13EDF6047C555302D2238D8CE11DF2424F1B66C2C5D238D0744DB679AF2890487031F9C0AEA1C4BB6FE9554EE528FDF1B05E5B256223B2F09215F3719F9C7CCC69DDF172D0D6234217FCC0037F18B93EF5389130B7A661E5C26E54214068BBCAFEA32A67818BD3075AD1F5C7E9CC3D1737FB28171BAF84DBB6612B7881C1A48E439CD03A92BF52225A2B38E6542E9F722BCE15A381B5753EA842763381CCAE83512B30511B32E5E8D80362149AD030AABA5F3A5798BB22AA7EC1B6D0F17903F4E22D840734AA85973F79A93FFB82A75C47C03D43D2F9CA02D03199BACEDDD4533A52566AFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```
 

FWIW, the prior prime was the 4,278,778th value larger than (2^256 - 1)*2^3840 + 2**256*(gamma) + (2^256 – 1) with gamma = floor(euler_constant*2^3584) that is one greater than a multiple of q.  But it turned out that the 3,793,825th such value also met all of the conditions.  Evidently the prior generation code attempted to optimize a bit too much and effectively tested only every second possible value.  It therefore missed the new value which was found with an odd number of iterations.  ☹

 
# q
The small prime q remains unchanged at 2^256-189.

 
# r
The new value of r (which is (p-1)/q) is as follows.

**Decimal:**
```
9019518416950528558373478086511232658951474842525520401496114928154304263969655687927867442562559311457926593510757267649063628681241064260953609180947464800958467390949485096429653122916928704841547265126247408167856620024815508684472819746384115369148322548696439327979752948311712506113890045287907335656308945630141969472484100558565879585476547782717283106837945923693806973017510492730838409381014701258202694245760602718602550739205297257940969992371799325870179746191672464736721424617639973324090288952006260483222894269928179970153634220390287255837625331668555933039199194619824375869291271098935000699785346405055160394688637074599519052655517388596327473273906029869030988064607361165803129718773877185415445291671089029845994683414682274353665003204293107284473196033588697845087556526514092678744031772226855409523354476737660407619436531080189837076164818131039104397776628128325247709678431023369197272126578394856752060591013812807437681624251867074769638052097737959472027002770963255207757153746376691827309573603635608169799503216990026029763868313819255248026666854405409059422844776556067163611304891154793770115766608153679099327786
```
 

**Hex:**
```
100000000000000000000000000000000000000000000000000000000000000BC93C467E37DB0C7A4D1BE3F810152CB56A1CECC3AF65CC0190C03DF34709B8AF6A64C0CEDCF2D559DA9D97F095C3076C686037619148D2C86C317102AFA2148031F04440AC0FF0C9A417A89212512E7607B2501DAA4D38A2C1410C4836149E2BDB8C8260E627C4646963EFFE9E16E495D48BD215C6D8EC9D1667657A2A1C8506F2113FFAD19A6B2BC7C45760456719183309F874BC9ACE570FFDA877AA2B23A2D6F291C1554CA2EB12F12CD009B8B8734A64AD51EB893BD891750B85162241D908F0C9709879758E7E8233EAB3BF2D6AB53AFA32AA153AD6682E5A0648897C9BE18A0D50BECE030C3432336AD9163E33F8E7DAF498F14BB2852AFFA814841EB18DD5F0E89516D557776285C16071D211194EE1C3F34642036AB886E3EC28882CE4003DEA335B4D935BAE4B58235B9FB2BAB713C8F705A1C7DE42220209D6BBCACC467318601565272E4A63E38E2499754AE493AC1A8E83469EEF35CA27C271BC792EEE21156E617B922EA8F713C22CF282DC5D6385BB12868EB781278FA0AB2A8958FCCB5FFE2E5C361FC174420122B0163CA4A46308C8C46C91EA7457C136A7D9FD4A7F529FD4A7F529FD4A7F529FD4A7F529FD4A7F529FD4A7F529FD4A7F52A
```
 
As required by the spec … and the thing that makes the search for p challenging, this value r is twice a prime.  (r will always be even, so r being twice a prime is the best that we can do.)

 
# g
Finally, the new value of g (which is 2^r mod p) is as follows.

**Decimal:**
```
119359756198641231858139651428439585561105914902686985078252796680474637856752833978884422594516170665312423393830118608408063594508087813277769835084746883589963798527237870817233369094387978405585759195339509768803496494994109693743279157584139079471178850751266233150727771094796709619646350222242437970473900636242584673413224137139139346254912172628651028694427789523683070264102332413084663100402635889283790741342401259356660761075766365672754329863241692760862540151023800163269173550320623249398630247531924855997863109776955214403044727497968354022277828136634059011708099779241302941071701051050378539485717425482151777277387633806111112178267035315726401285294598397677116389893642725498831127977915200359151833767358091365292230363248410124916825814514852703770457024102738694375502049388804979035628232209959549199366986471874840784466132903083308458356458177839111623113116525230200791649979270165318729763550486200224695556789081331596212761936863634467236301450039399776963661755684863012396788149479256016157814129329192490798309248914535389650594573156725696657302152874510063002532052622638033113978672254680147128450265983503193865576932419282003012093526302631221491418211528781074474515924597472841036553107847
```

**Hex:**
```
1D41E49C477E15EAEEF0C5E4AC08D4A46C268CD3424FC01D13769BDB43673218587BC86C4C1448D006A03699F3ABAE5FEB19E296F5D143CC5E4A3FC89088C9F4523D166EE3AE9D5FB03C0BDD77ADD5C017F6C55E2EC92C226FEF5C6C1DF2E7C36D90E7EAADE098241D3409983BCCD2B5379E9391FBC62F9F8D939D1208B160367C134264122189595EC85C8CDBE5F9D307F46912C04932F8C16815A76B4682BD6BDC0ED52B00D8D30F59C731D5A7FFAE8165D53CF96649AAC2B743DA56F14F19DACC5236F29B1AB9F9BEFC69697293D5DEAD8B5BF5DE9BAB6DE67C45719E56344A3CBDF3609824B1B578E34EAEB6DD3190AB3571D6D671C512282C1DA7BD36B4251D2584FADEA80B9E141423074DD9B5FB83ACBDEAD4C87A58FFF517F977A83080370A3B0CF98A1BC2978C47AAC29611FD6C40E2F9875C35D50443A9AA3F49611DCD3A0D6FF3CB3FACF31471BDB61860B92C594D4E46569BB39FEEADFF1FD64C836A6D6DB85C6BA7241766B7AB56BF739633B054147F7170921412E948D9E47402D15BB1C257318612C121C36B80EB8433C08E7D0B7149E3AB0A8735A92EDCE8FF943E28A2DCEACFCC69EC318909CB047BE1C5858844B5AD44F22EEB289E4CC554F7A5E2F3DEA026877FF92851816071CE028EB868D965CCB2D2295A8C55BD1C070B39B09AE06B37D29343B9D8997DC244C468B980970731736EE018BBADB987
```

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #329
Fixes #76

### Description
- Adjust STANDARD primes to be used in production cases
- Add other test prime cases (Miniscule, Tiny, Small, and Medium)
- Migrate ElectionConstants to new file and isolate from Serialized class to avoid circular dependency. The eventual move is to remove this class. 
- Add in small_prime, large_prime, cofactor and generator in replacement of Q, P, R, G constants to handle live reloading of the constants with environment variables. This allows better testing with using the environment variables. 

### Testing
Existing test suite is consistent enough to test the use of the new primes. 
There is a new BaseTestCase underneath all the tests.

**Before**
Screen Shot 2021-07-14 at 3 12 02 PM

**After (Medium Test Primes)**
Screen Shot 2021-07-16 at 5 09 04 PM


src/electionguard/constants.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+
+    option_map = {
+        PrimeOption.Standard: STANDARD_CONSTANTS,
+        PrimeOption.TestOnly: MEDIUM_TEST_CONSTANTS,
+    }
+    return option_map.get(option) or STANDARD_CONSTANTS
+
+
+get_large_prime = lambda: get_constants().large_prime
```
keithrfung, line 75: These functions were added as opposed to constants to ensure the testing suite can stay flexible. The environment variables can be changed quickly on the tests allowing for different testing cases which has a lot of potential weight long term. 

Before Changes
Screen Shot 2021-07-14 at 3 17 42 PM

After Changes
After changes

AddressXception (approved):

src/electionguard/constants.py
```diff
@@ -0,0 +1,81 @@
...
+    return ElectionConstants(
+        large_prime,
+        small_prime,
+        cofactor,
+        generator,
+    )
+
+
+# pylint: disable=line-too-long
+STANDARD_CONSTANTS = create_constants(
```
AddressXception, line 38: how do you feel about adding a comment here that includes a link to the spec where primes are specified?

AddressXception (changes requested):

src/electionguard/constants.py
```diff
@@ -0,0 +1,81 @@
...
+)
+
+# TEST ONLY
+# These constants serve as sets of primes for future developers
+# Currently, all the sets are all valid but may break certain tests
+# As tests adapt, these constants can be used to speed up tests
+EXTRA_SMALL_TEST_CONSTANTS = create_constants(157, 13, 12, 16)
+SMALL_TEST_CONSTANTS = create_constants(503, 251, 2, 5)
```
danwallach, line 50: I don't really see the point of teeny tiny tests like this. The "medium" test constants will run incredibly fast and will thus be more likely to uncover bugs, if they exist. (There are also some assumptions, in the tests, were we assume that collisions never ever happen, which are perfectly reasonable assumptions when we're dealing with 256-bit numbers that might as well be drawn randomly, but when I've experimentally tried these smaller constants, some tests fail. Those tests are still absolutely correct, they just shouldn't run for the tiny constants.)

src/electionguard/constants.py
danwallach: Good to see this. I would have been nervous about going from constants to functions.

src/electionguard/constants.py
keithrfung: While I agree with that it is not good practice to use the tiny constants for the entire test suite. 

Having a couple sets of constants available to the public can still be helpful.The suite as it stands still takes 2-3 min to run. These constants could be used for a single test to help speed up with certain cases or if more tests are added. Perhaps even running a lighter test suite for PR's instead of Releases.

danwallach (commented):
Overall, I like how this change works and how you can just tweak an environment variable to get the new behavior. If I had to pick nits, I'd say I don't like `small_prime` and `large_prime`, since everywhere else we're using terms like `ElementModP` and `ElementModQ`.  These terms should all be consistent. 

A few things worth paying attention to, although these probably need to be a separate feature request:

- When serializing and deserializing, there needs to be some logic to note which election constants were in use. My gut says that we should simply fail if serializing anything smaller than the production constants.
- We might want to add a field to `ElementModP` and `ElementModQ` that indicates which of the election constants was in use. While this makes each of those wrapper classes a little bit larger, I've shoved a lot more in there for the `electionguard-edu` variants, and there's no impact on performance. This might then be all the support you need for the serialization code to reject anything that's not "full strength". This might be as simple as:

```python
class ElementModQ(NamedTuple):
    """An element of the smaller `mod q` space, i.e., in [0, Q), where Q is a 256-bit prime."""

    elem: mpz
    prime_option: PrimeOption

    def to_hex(self) -> str:
        """
        Convert from the element to the hex representation of bytes.
        This is preferable to directly accessing `elem`, whose representation might change.
        """

        if self.prime_option != PrimeOption.Standard:
            massiveEarthShatteringKaboom()
        h = format(self.elem, "02X")
        if len(h) % 2:
            h = "0" + h
        return h
```

src/electionguard/constants.py
keithrfung: Added this in markdown

AddressXception (approved):

PR #396, opened 2021-07-23
Make discrete_log asynchronous

### LINKED ISSUE #393: Remove asyncio.run call in dlog.py
## Bug Report

**Expected Behavior**
When executing a function that calls the discrete log lookup from a running event loop, it should process the dlog lookup and allow the function to continue.

**Actual Behavior**
the dlog throws with `RuntimeError('asyncio.run() cannot be called from a running event loop')`

AddressXception:
### Issue
*Link your PR to an issue*

Fixes #393

### Description
Make the dlog async/await and propagate through the rest of the API's.  This is not an ideal solution for the long term but it unblocks the issue in the short term allowing the code to be called from inside a running event loop.

Further work to fix this correctly is tracked in #397 https://github.com/microsoft/electionguard-python/issues/397

### Testing
1. run the tests

### DIFF
--- a/docs/4_Decrypt_Tally.md
+++ b/docs/4_Decrypt_Tally.md
@@ -1,22 +1,23 @@
 # Decryption
 
-At the conclusion of voting, all of the cast ballots are published in their encrypted form in the election record together with the proofs that the ballots are well-formed.  Additionally, all of the encryptions of each option are homomorphically-combined to form an encryption of the total number of times that each option was selected.  The homomorphically-combined encryptions are decrypted to generate the election tally.  Individual cast ballots are not decrypted.  Individual spoiled ballots are decrypted and the plaintext values are published along with the encrypted representations and the proofs.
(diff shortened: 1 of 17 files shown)


danwallach:
Propagating all that async stuff everywhere is a real mess for using ElectionGuard as a library. Much better to provide some sort of bridge so you can call an async version or a synchronous version. If you're making changes in elgamal.py, that's really bad.

keithrfung:
@danwallach 
The bigger issue is the discrete log function`dlog` which is sitting underneath is already asynchronous but is wrapped in synchronous calls. Your point is relevant, but it's already present with the dlog as it is currently designed. 

This change highlights the clear asynchronous chain that's dependent on dlog. 

Consider for a moment, that an `async` and `sync` set would still require making a truly non async dlog. The core root issue. If this could be completed, then there would be no need for these changes.

danwallach:
The issue isn't synchronous vs asynchronous. The issue is that there needs to be a write-lock around the results cache. If async pollution is the problem, the solution is to find a different way to implement the write lock.

AddressXception:
agreed, what as missing here was this ticket to track the follow-on work as part of a larger epic.

#397

AddressXception:
replaced by #405

Issue #397, opened 2021-07-26
Add precompute runtime optimizations


AddressXception:
## Feature Request

**Description**
As a user of the library i would like to precompute data tables used in ballot encryption and decryption that do not rely on sensitive data such as the nonces so that i can optimize the real-time perceived performance when executing encryption and decryption operations

specifically, a number of the exponentiations calculated whne encrypting a ballot are not associated with the nonce's and therefore can be precomputed in a lookup table.

similarly the discrete log lookup table used when decrypting is currently precomputed using a global cache and async function calls.

The implementation should use a consistent and reproducable pattern similar to other segments of the code which follow a functional pattern at the lowest level and include a convenience wrapper that can manage objects and global state.

keithrfung:
I believe this is a duplicate of #410

PR #405, opened 2021-08-02
Modify discrete log to support sync and async

### LINKED ISSUE #393: Remove asyncio.run call in dlog.py

AddressXception:

### Issue
*Link your PR to an issue*

Fixes #393 

### Description
replaces [PR #396](https://github.com/microsoft/electionguard-python/pull/396).  Make the discrete log sync or async.  the methods use the synchronous version.  Further work to load the dlog from a cache is tracked in #397 

If necessary, async versions of the decryption logic may need to be added in the future.

### Testing
1. run the tests


src/electionguard/discrete_log.py
```diff
@@ -0,0 +1,100 @@
...
+
+import asyncio
+from typing import Dict, Tuple
+
+from electionguard.singleton import Singleton
+
+from .group import G, ElementModP, ONE_MOD_P, mult_p, int_to_p_unchecked
+
+DLOG_CACHE = Dict[ElementModP, int]
+DLOG_MAX = 1_000_000
```
AddressXception, line 12: i set the default here to 1m but it could be something else.  it was 100m in dan's example

src/electionguard/discrete_log.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def discrete_log(element: ElementModP, cache: DLOG_CACHE) -> Tuple[int, DLOG_CACHE]:
```
AddressXception, line 15: the functional methods now expect a cache object and return a cache object with the lookup result, making it easier for a consumer to define their own caching strategy

src/electionguard/discrete_log.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    if element in cache:
+        return (cache[element], cache)
+
+    cache = compute_discrete_log_cache(element, cache)
+    return (cache[element], cache)
+
+
+async def discrete_log_async(
```
AddressXception, line 35: now if a caller wants to explicitly have thread safety they can call this method with their own lock and event loop

src/electionguard/discrete_log.py
```diff
@@ -0,0 +1,100 @@
...
+
+def compute_discrete_log_cache(
+    element: ElementModP, cache: DLOG_CACHE
+) -> Dict[ElementModP, int]:
+    """
+    Compute a discrete log cache up to the specified element.
+    """
+    if not cache:
+        cache = {ONE_MOD_P: 0}
+    max_element = list(cache)[-1]
```
AddressXception, line 70: python 3.7+ now guarantees the insertion order, so no need for `OrderedDict`

src/electionguard/discrete_log.py
```diff
@@ -0,0 +1,100 @@
...
+    exponent = cache[max_element]
+
+    g = int_to_p_unchecked(G)
```
keithrfung, line 73: I removed this in my other PR. This works but just denoting this line will need to be fixed later.

danwallach (commented):
It seems like overkill to go from global state to state within a class instance, but for most common usages of `discrete_log`, there won't be a practical difference.

The decision to raise an exception is a bit more interesting, because now you need to audit all the places that ask for a discrete log and make sure that they're prepared to catch that exception or it's okay for them to fail. 

In my version, I just had it return `Optional[int]`, which caused more pain initially, but the payoff was that `mypy` helped me find each and every place that was depending on that value succeeding, and then add more error handling where needed.

BTW, as to having the maximum value be 1 million or 100 million, you want this value to be big enough to encompass the homomorphic tally of the biggest election that ElectionGuard will ever handle. In L.A. County, California (which is probably the biggest single election jurisdiction in the U.S.), there were somewhere north of 4 million votes cast in November 2020. So, I picked 100 million to be really, really bigger than that, but still small enough that if you computed the whole thing by accident, you wouldn't run for a huge amount of time only to fail.

src/electionguard/discrete_log.py
```diff
@@ -0,0 +1,100 @@
...
+    while element != max_element:
+        exponent = exponent + 1
+        if exponent > DLOG_MAX:
+            raise Exception("size is larger than max.")
```
keithrfung, line 77: This should be throwing a `ValueError` . 

https://docs.python.org/3/library/exceptions.html#ValueError

Possibly a KeyError or IndexError but not a generic Exception.

danwallach (commented):
Also, if you really want to go the o-o route, I'd kill off the top-level functions and inline them into the `DiscreteLog` class, since anybody outside of this file shouldn't have to know about which mutex or which cache to use.

src/electionguard/discrete_log.py
```diff
@@ -0,0 +1,100 @@
...
+    Note: *this function is thread-safe*. For the best possible performance,
+    pre-compute the discrete log of a number you expect to have the biggest
+    exponent you'll ever see. After that, the cache will be fully loaded,
+    and every call will be nothing more than a dictionary lookup.
+    """
+    if element in cache:
+        return (cache[element], cache)
+
+    async with mutex:
+        if element in cache:
```
keithrfung, line 55: Why is this line in here twice?

src/electionguard/discrete_log.py
keithrfung: Does this method have to exist in the library though? 

This is available now by extending just the existing functions. With the changes, this reduces the need to include this anymore. And if it is used, it should be rewritten to show what's happening. 

Why not just
```
async with mutex
      return discrete_log(...)
```

src/electionguard/discrete_log.py
keithrfung: This should do something to what happens in the primes. This could be set with environment variables. Default to this but be overridable via environment.

tests/property/test_discrete_log.py
```diff
@@ -0,0 +1,99 @@
...
+    g_pow_p,
+    int_to_q,
+    int_to_p_unchecked,
+    int_to_q_unchecked,
+    P,
+)
+from electionguard.utils import get_optional
+
+
+# simpler implementation of discrete_log, only meant for comparison testing of the caching version
```
keithrfung, line 25: This should be in the function definition, not just a comment line.

src/electionguard/elgamal.py
```diff
@@ -48,7 +48,7 @@ def decrypt_known_product(self, product: ElementModP) -> int:
         :param product: The known product (blinding factor).
         :return: An exponentially encoded plaintext message.
         """
-        return discrete_log(mult_p(self.data, mult_inv_p(product)))
+        return DiscreteLog().discrete_log(mult_p(self.data, mult_inv_p(product)))
```
keithrfung, line 51: Doesn't feel super clean, but I don't have something to replace it with besides a function.

keithrfung (changes requested):

src/electionguard/discrete_log.py
danwallach: This is exceptionally important. Before you acquire the lock, a second thread could come through and change the cache. That's why you check it again after you acquire the lock. This was a bug in my original code, which I only found after rewriting this in Kotlin.

src/electionguard/discrete_log.py
danwallach: Or, see comments above, the function should return `Optional[int]` rather than `int` with the possibility of an unchecked exception blowing things up downstream.

danwallach: Also, the error message needs to be more descriptive, like "discrete_log failure; possibly caused by bad data encryption".

src/electionguard/discrete_log.py
AddressXception: value can change between entry to and after the mutex is acquired

AddressXception:
> Also, if you really want to go the o-o route, I'd kill off the top-level functions and inline them into the `DiscreteLog` class, since anybody outside of this file shouldn't have to know about which mutex or which cache to use.

the idea is to have flexibility, allowing a consumer to use the functional low level process or the higher order "stateful" process. that's consistent with most of the other api's we have (e.g. encryptSelection, encryptContest, encryptBallot)

keithrfung (dismissed):

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "electionguard"
-version = "1.2.2"
+version = "1.2.3"
```
keithrfung, line 3: Good 👍

keithrfung (approved):

PR #415, opened 2021-08-06
Crypto Hash using Sequence Order

### LINKED ISSUE #350: CiphertextBallot crypto_hash_with does not enforce ordering
## Bug Report

**Expected Behavior**

When generating crypto hashes on ciphertext ballots, the hashes should always be created in the same order

in [ballot.py#CiphertextBallot.crypto_hash_with](https://github.com/microsoft/electionguard-python/blob/main/src/electionguard/ballot.py#L743):

```python
contest_hashes = [contest.crypto_hash for contest in self.contests]
```
and in [ballot.py#_ciphertext_ballot_context_crypto_hash](https://github.com/microsoft/electionguard-python/blob/71e8150e3f622de1f1c6cb177bd7b7e954150943/src/electionguard/ballot.py#L540)

```python
selection_hashes = [selection.crypto_hash for selection in ballot_selections]
```

**Actual Behavior**

we assume the order is correct, however if the user generated the encryptions using the functional interfaces (and not the `encrypt_ballot` function) there is no way to guarantee.

**Proposed Solution**

Add the `sequence_order` of each `SelectionDescription` and each `ContestDescription` from the `InternalManifest` to the expected required values of each object constructor and make function.  This solution retains the separation of the ballot from the manifest.:

e.g.:

```python
def make_ciphertext_ballot_selection(
    object_id: str,
    description_hash: ElementModQ,
    ciphertext: ElGamalCiphertext,
    elgamal_public_key: ElementModP,
    crypto_extended_base_hash: ElementModQ,
    proof_seed: ElementModQ,
    # **** DEV: add the sequence order ****
    sequence_order: int,
    selection_representation: int,
    is_placeholder_selection: bool = False,
    nonce: Optional[ElementModQ] = None,
    crypto_hash: Optional[ElementModQ] = None,
    proof: Optional[DisjunctiveChaumPedersenProof] = None,
    extended_data: Optional[ElGamalCiphertext] = None,
) -> CiphertextBallotSelection:
```

and:

```python
@runtime_checkable
class CiphertextSelection(Protocol):

    object_id: str

    # **** DEV: add the sequence order ****
    sequence_order: int

    description_hash: ElementModQ
    """The SelectionDescription hash"""

    ciphertext: ElGamalCiphertext
    """The encrypted representation of the selection"""
```

then something like:

```python
selection_hashes = [
    selection.crypto_hash for selection in sorted(
        ballot_selections, key=lambda c: c.sequence_order
    )]
```

and repeat for contests, etc.  This code change will ripple through a number of files including `encrypt.py`.

also, `ballot.py#_ciphertext_ballot_context_crypto_hash` should probably be renamed to `ballot.py#_ciphertext_ballot_contest_crypto_hash` (**contest** and not _context_)

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #350 

### Description
The crypto hash did not have obvious problems but did not correctly ensure ordering. The item present to check for this was the sequence order. This resolves this problem by ensuring the ballot maintains the ordering allowing quick sorting. This is important to have on the ballot to ensure that without the manifest the ballot contests and selections can still be correctly sorted. This also helps solve equality problems that could present themselves and simplify things on the compact ballot. 

The approach was to add the sequence_order slowly into the ciphertext pieces and then allow this to ripple into the plaintext. A method was added to sort to ensure that the sort function was minimized. It currently exists only in a singular location. 

### Testing
The main tests to review are the tests around decryption and tally. This being said the existing test suite has coverage on this pull request.

### DIFF
--- a/src/electionguard/ballot.py
+++ b/src/electionguard/ballot.py
@@ -10,7 +10,11 @@ from .chaum_pedersen import (
     make_constant_chaum_pedersen,
     make_disjunctive_chaum_pedersen,
 )
-from .election_object_base import ElectionObjectBase
+from .election_object_base import (
+    ElectionObjectBase,
+    OrderedObjectBase,
+    sequence_order_sort,
+)
 from .elgamal import ElGamalCiphertext, elgamal_add
 from .group import add_q, ElementModP, ElementModQ, ZERO_MOD_Q
 from .hash import CryptoHashCheckable, hash_elems
@@ -43,7 +47,7 @@ class ExtendedData:
 
 
 @dataclass(unsafe_hash=True)
-class PlaintextBallotSelection(ElectionObjectBase):
+class PlaintextBallotSelection(OrderedObjectBase):
     """
     A BallotSelection represents an individual selection on a ballot.
 
@@ -112,6 +116,9 @@ class CiphertextSelection(Protocol):
 
     object_id: str
 
+    sequence_order: int
+    """Order the selection."""
+
     description_hash: ElementModQ
     """The SelectionDescription hash"""
 
@@ -121,7 +128,7 @@ class CiphertextSelection(Protocol):
 
 @dataclass(eq=True, unsafe_hash=True)
 class CiphertextBallotSelection(
-    ElectionObjectBase, CiphertextSelection, CryptoHashCheckable
+    OrderedObjectBase, CiphertextSelection, CryptoHashCheckable
 ):
     """
     A CiphertextBallotSelection represents an individual encrypted selection on a ballot.
@@ -238,6 +245,7 @@ def _ciphertext_ballot_selection_crypto_hash_with(
 
 def make_ciphertext_ballot_selection(
     object_id: str,
+    sequence_order: int,
     description_hash: ElementModQ,
     ciphertext: ElGamalCiphertext,
     elgamal_public_key: ElementModP,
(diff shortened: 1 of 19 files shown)


src/electionguard/ballot.py
```diff
@@ -19,15 +23,13 @@
...
 ) -> bool:
     """
     We want to compare lists of election objects as if they're sets. We fake this by first
-    sorting them on their object ids, then using regular list comparison.
+    sorting them on their sequence order, then using regular list comparison.
     """
-    return sorted(list1, key=lambda x: x.object_id) == sorted(
-        list2, key=lambda x: x.object_id
-    )
+    return sequence_order_sort(list1) == sequence_order_sort(list2)
```
keithrfung, line 32: This change could go unnoticed so calling it out. This modifies the sort of equality to rely on `sequence_order` instead of `object_id`. Considering `sequence_order` represents ordering this makes more sense and is consistent with how the items should be sorted.

src/electionguard/ballot_compact.py
```diff
@@ -154,29 +155,30 @@ def _get_plaintext_contests(
     )
 
     contests: List[PlaintextBallotContest] = []
-    for manifest_contest in sorted(
-        internal_manifest.contests, key=lambda c: c.sequence_order
```
keithrfung, line 158: Any location where there was a lambda sort it was exchanged to rely exclusively on the new `sequence_order_sort()`

src/electionguard/election_object_base.py
```diff
@@ -1,8 +1,32 @@
...
 
 @dataclass
 class ElectionObjectBase:
     """A base object to derive other election objects identifiable by object_id."""
 
     object_id: str
+
+
+@dataclass
+class OrderedObjectBase(ElectionObjectBase):
```
keithrfung, line 15: This new base helps identify which objects can be ordered. It also ensures that the first two elements in the constructor are always `object_id` then `sequence_order` to avoid confusion. This did cause a ripple effect through a couple other classes.

AddressXception (approved):

PR #416, opened 2021-08-10
Update init using mkinit cli

### LINKED ISSUE #395: Add `__all__` to electionguard module
## Feature Request

**Description**
`__all__` helps define all the exported methods and classes. A key issue is that it's a string but this is ideal to keep this package as usable as possible. This would be for exclusively the `electionguard` `init.py` 
https://docs.python.org/3/tutorial/modules.html

Something similar to this would be ideal:
https://github.com/samuelcolvin/pydantic/blob/4a54f393ad20ee91b51cd7a49ec46771ba4f8a18/pydantic/__init__.py#L22

RiyaJohn:
Fixes #395 

### Description
Auto updated init of electionguard using mkinit cli, more details at https://github.com/Erotemic/mkinit

### Testing
make test passed


keithrfung:
I like this but if we use `mkinit` we should ensure it's in the workflow in case anyone adds a method.

keithrfung (changes requested):
I like the approach on this, but I think we need a bit more.

- [ ] Fix linting**
This should like get fixed by using `mkinit` black formatting using the `--black` option on the command.
`poetry run mkinit --write --black src/electionguard`

- [ ] Add dependency to dev dependencies
`poetry add --dev mkinit`

- [ ] Add command to run mkinit**
Add mkinit command `make auto-lint` to MakeFile.
This should just autolint the `src/electionguard` with the `--black` option.
`poetry run mkinit --write --black src/electionguard`

- [ ] Add lint check for mkinit**
Implement a quick check to the diff to see if there is any changes. 

Here's a quick coded example. Needs a bit more. 
```
...
@echo 3. Mkinit Init Formatting
 poetry run mkinit --diff --black  src/electionguard  | grep -q '+' && echo "fails" || echo "success"
 @ echo 4. Mypy Static Typing 
 ...
```

RiyaJohn:
> I like the approach on this, but I think we need a bit more. If we're auto generating this, we want to ensure future developers have the `__init__.py` checked and can easily have it autocreated/autoformatted.
> 
> * [ ]  Fix linting
>   This should like get fixed by using `mkinit` black formatting using the `--black` option on the command.
>   `poetry run mkinit --write --black src/electionguard`
> * [ ]  Add dependency to dev dependencies
>   Run `poetry add --dev mkinit`
> * [ ]  Add lint command to run mkinit (`auto-lint` in `MakeFile`)
>   This should just autolint the `src/electionguard` with the `--black` option.
>   `poetry run mkinit --write --black src/electionguard`
> * [ ]  Add lint check for mkinit (`lint` in `MakeFile`)
>   Implement a quick check to the diff to see if there is any changes. Here's a quick coded example where I take advantage of the `diff` function and look for any changes. Needs a bit more.
> 
> ```
> ...
> @echo 3. Mkinit Init Formatting
>  poetry run mkinit --diff --black  src/electionguard  | grep -q '+' && echo "fails" || echo "success"
>  @ echo 4. Mypy Static Typing 
>  ...
> ```

I had initially considered adding this to auto generate every time. But though I used mkinit to generate the file, I had to manually fix some issues like in a file a local variable was also imported in the auto generate that had to be removed manually. Also, some constants were reimported from multiple files that had to be removed manually: 
```
src/electionguard/__init__.py:90:0: W0404: Reimport 'ELECTION_PUBLIC_KEY' (imported line 83) (reimported)
src/electionguard/__init__.py:98:0: W0404: Reimport 'ELECTION_PUBLIC_KEY' (imported line 83) (reimported)
src/electionguard/__init__.py:98:0: W0404: Reimport 'GUARDIAN_PUBLIC_KEY' (imported line 90) (reimported)
src/electionguard/__init__.py:116:0: W0404: Reimport 'ELECTION_PUBLIC_KEY' (imported line 83) (reimported)
src/electionguard/__init__.py:127:0: W0404: Reimport 'discrete_log' (imported line 15) (reimported)
src/electionguard/__init__.py:167:0: W0404: Reimport 'ELECTION_PUBLIC_KEY' (imported line 83) (reimported)
```
Hence I dropped the idea of using this in make, but let me know what you think around this.

keithrfung:
@RiyaJohn Hmm are these issues that actually exist like referencing the same import in different ways where we could rely on the autogenerate with some fixes? I think autogenerate is definitely what we should try to lean on.

RiyaJohn:
> @RiyaJohn Hmm are these issues that actually exist like referencing the same import in different ways where we could rely on the autogenerate with some fixes? I think autogenerate is definitely what we should try to lean on.

Made some fixes in the code to autogenerate mkinit. But not sure how to fix the multiple imports of the 2 global variables

RiyaJohn:
Using --black with mkinit cmd wasnt effective. Hence added it as the next command to reformat init after autogeneration.
_Implement a quick check to the diff to see if there is any changes. Here's a quick coded example where I take advantage of the diff function and look for any changes. Needs a bit more._
 For similar reasons, did not add it here, because init file gets completely reformatted when black command is applied and hence this will always show as "fails"

keithrfung:
@RiyaJohn since there were changes to be made to the base. I pulled your branch and made a new PR. Take a look at that. I kept your commits so that your contribution could still be seen. See #424

PR #451, opened 2021-09-29
verifier issue 426

### LINKED ISSUE #426: 📦 Create an Election Verifier folder to start module
## Feature Request

**Description**
This package is to create an `electionguard_verifier` package. This should be similar to how the package for `electionguard` is setup. This issue is to initially setup the package so future cards can implement the verify tasks. 

- [x] Create `electionguard_verify` folder
- [x] Create a `verify_ballot` method that takes a ballot and always returns true but is marked to indicate it's a temporary method. 
- [x] Setup auto-init to handle `electionguard_verify`
- [x] Setup linting for the package
- [x] Setup a corresponding test project with a test that loads a ballot and passes it to the `verify_ballot` function. 
- [x] Create any necessary make adjustments for builds

This replaces #78 . 

----------------------

```py
def verify_ballot(ballot: CiphertextBallot) -> Verification:
        return Verification(True)
```

```py
@dataclass
class Verification
      verified: bool
      message: Optional[str]
```

cryptosidh:
### Issue

Fixes #426 

### Description
Created `electionguard_verifier` package with a temporary `verify_ballot` method that always returns true and a corresponding test project that asserts `verify_ballot` returns True on an encrypted ballot.

### Testing
Run `test_verify_ballot`, which is included in property-tests.


src/electionguard_tools/__init__.py
```diff
@@ -35,6 +34,7 @@
     RESULTS_DIR,
     T,
     TALLY_FILE_NAME,
+    TallyCeremonyOrchestrator,
```
keithrfung, line 37: Odd that this came up. Since this is autogenerated. Leaving this off the review

keithrfung (approved):
A good step in the right direction. We're off to making verifiers.

PR #456, opened 2021-10-04
Single source version from pyproject.toml

### LINKED ISSUE #422: ♻️ Single source the version from pyproject.toml
# Feature Request

`__version__` should be present in the __init__.py of `electionguard` and `testelectionguard`

This shouldn't be duplicated and should run off the pyproject.toml version.
Here is an [example](https://gist.github.com/benkehoe/066a73903e84576a8d6d911cfedc2df6).

ronnie-llamado:

### Issue
Fixes #422

### Description
- Add `__version__` that sources from `pyproject.toml` to all `__init__.py` in subpackages
- Add XML tags to contain `mkinit` generated code
- Add `importlib-metadata` to poetry when version is <3.8
- Note: Ignored second import in all `__init__.py` to prevent mypy error - [`no-redef`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-each-name-is-defined-once-no-redef)

### Testing
```python
>>> import electionguard, electionguard_tools, electionguard_verify
>>> electionguard.__version__
'1.3.0'
>>> electionguard_tools.__version__
'1.3.0'
>>> electionguard_verify.__version__
'1.3.0'
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -222,9 +222,9 @@ python-versions = ">=3.5"
 
 [[package]]
 name = "importlib-metadata"
-version = "4.6.3"
+version = "4.8.1"
 description = "Read metadata from Python packages"
-category = "dev"
+category = "main"
 optional = false
 python-versions = ">=3.6"
 
@@ -985,7 +985,7 @@ tests = ["codecov", "scikit-build", "cmake", "ninja", "pybind11", "pytest", "pyt
 name = "zipp"
 version = "3.5.0"
 description = "Backport of pathlib-compatible object wrapper for zip files"
-category = "dev"
+category = "main"
 optional = false
 python-versions = ">=3.6"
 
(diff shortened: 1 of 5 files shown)


keithrfung:
@ronnie-llamado Super cool work. Love that you thought ahead about mkinit and made sure it would still work. I'm going to double check with the crew about python <3.8 support because I think this code would be a lot cleaner without it, but solid pr. Keep it up!

For any others reviewing:
> For more fine grained control, you can specify XML-like # <AUTOGEN_INIT> and # </AUTOGEN_INIT> comments around the volitle area. If specified only the area between these tags will be overwritten.

Source: https://mkinit.readthedocs.io/en/latest/autoapi/mkinit/

keithrfung (changes requested):
@ronnie-llamado I apparently mispoke. We don't have to worry about pre 3.8 so this can allow you to clean up your code and remove the exception for coverage before that version.

PR #457, opened 2021-10-06
add verify_ballot method and edit accompanying test

### LINKED ISSUE #430: ✅. Verifier Package: Verify Ballot
## Feature Request

**Description**
Create a verify method to verify the ballot is well formed, the proofs work, and that the contests and selections are valid.
Method should be added to the Verifier package from #426. 

- [ ] Create `verify_ballot` method
- [ ] Create accompanying test

PradyumnaKrishna:
### Issue

Fixes #430 

### Description
Created a `verfiy_ballot` method that checks the encryption of ballot, the proofs work, and that the contests and selections are valid. Edited their accompanying tests.

### Testing
Run `test_verify_ballot`, which is included in property-tests.

### DIFF
--- a/src/electionguard_verify/verify.py
+++ b/src/electionguard_verify/verify.py
@@ -4,6 +4,10 @@ from typing import Optional
 from electionguard.ballot import (
     CiphertextBallot,
 )
+from electionguard.election import CiphertextElectionContext
+from electionguard.manifest import (
+    InternalManifest,
+)
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


src/electionguard_verify/verify.py
```diff
@@ -17,13 +21,23 @@ class Verification:
     message: Optional[str]
 
 
-def verify_ballot(ballot: CiphertextBallot) -> Verification:
+def verify_ballot(
+    ballot: CiphertextBallot,
+    internal_manifest: InternalManifest,
```
keithrfung, line 26: In this particular case, we do not want to send the `InternalManifest` but the `Manifest` for the method signature. 

Luckily one can move between the two. But this method should instead take a `Manifest` as a parameter.

tests/property/test_verify.py
```diff
@@ -43,8 +43,8 @@ def test_verify_ballot(self, keypair: ElGamalKeyPair):
         self.assertIsNotNone(encrypted_ballot)
 
         # Act
-        verification = verify_ballot(encrypted_ballot)
+        verification = verify_ballot(encrypted_ballot, election, context)
```
keithrfung, line 46: Minor note. This variable should be `manifest` not election.

keithrfung (changes requested):

PR #458, opened 2021-10-06
Workaround for Python 3.9.7 operation

### LINKED ISSUE #440: 🐞  Python 3.9.7 causes tests to fail

### Current Behavior

Python 3.9.7 was found by the CI to cause issues with the existing test suite. 

### Expected Behavior

Tests should run as 3.9.5 and successfully pass. 

### Steps To Reproduce

1. Set environment to 3.9.7
This can be done with `poetry env use 3.9.7` and `pyenv local 3.9.7`

2. Run `make tests`

### Environment

```markdown
- OS: Mac, Windows, Linux
```

### Anything else?

This workflow shows the bugs if you wish to see the problems before starting
https://github.com/microsoft/electionguard-python/runs/3655735644?check_suite_focus=true

ronnie-llamado:

### Issue
Fixes #440

### Description
- Explicitly defines `__init__` for `Manifest`,`InternationalizedText`,`ContactInformation`, and `Language`
  - Fails if `init=False` is defined in dataclass

### Testing
1. Set environment to 3.9.7
2. `make test`

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -11,7 +11,7 @@ on:
     - cron: "30 23 * * 1" # 2330 UTC Every Monday
 
 env:
-  PYTHON_VERSION: "3.9.5"
+  PYTHON_VERSION: 3.9
   POETRY_PATH: "$HOME/.poetry/bin"
 
 jobs:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,7 +5,7 @@ on:
     types: [closed]
 
 env:
-  PYTHON_VERSION: "3.9.5"
+  PYTHON_VERSION: 3.9
   POETRY_PATH: "$HOME/.poetry/bin"
 
 jobs:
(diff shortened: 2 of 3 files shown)


keithrfung:
@ronnie-llamado It's a creative fix but I believe this works. 

I believe `AnnotatedString` doesn't throw because of the ordering of the elements. `AnnotatedString` has only optional parameters where all the others have both. 

I will make an issue to remove these in the future. 

The README and any other documentation can now say 3.9 instead of a specific version on the changes.

ronnie-llamado:
> I believe `AnnotatedString` doesn't throw because of the ordering of the elements. `AnnotatedString` has only optional parameters where all the others have both.

I had the same thought at one point, but `ContactInformation` also only has optional parameters. 

It actually looks like `AnnotatedString` is not being instantiated in the testing, so no errors are being thrown. When generating `ContactInformation`, emails are passed in straight from `hypothesis` and no phone numbers are passed in. These two are expected to be of type `AnnotatedString`.

https://github.com/microsoft/electionguard-python/blob/f3e2f09503cab2214706240404ca70d5f7a925b0/src/electionguard_tools/strategies/election.py#L178-L185

PR #460, opened 2021-10-08
Modify encrypt_ballot to pass on should_verify_proofs

### LINKED ISSUE #459: 🐞  should_verify_proofs cannot actually be overridden from ballot level down

### Current Behavior

Although `should_verify_proofs` is set to `False`, it is not passed into the other functions so proofs will still be verified in the underlying contests and selections. 

https://github.com/microsoft/electionguard-python/blob/a29590262260db0e9ee0b080c7ed75d278c1478c/src/electionguard/encrypt.py#L161
https://github.com/microsoft/electionguard-python/blob/f3e2f09503cab2214706240404ca70d5f7a925b0/src/electionguard/encrypt.py#L243
https://github.com/microsoft/electionguard-python/blob/f3e2f09503cab2214706240404ca70d5f7a925b0/src/electionguard/encrypt.py#L410

### Expected Behavior

The `should_verify_proofs` should be passed from Ballot to Contest and from Contest to Selection

### Steps To Reproduce

1. Call `encrypt_ballot` with `should_verify_proofs` set to `False`

### Environment

```markdown
- OS: Any
```

### Anything else?

_No response_

shepherdjay:

### Issue

Fixes #459

### Description

Cascades the should_verify_proofs down the call stack

### Testing

A test is provided, tried to match the unittest style. Uses patch with a sideffect to original function to track calls.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -16,7 +16,7 @@ python-versions = "*"
 
 [[package]]
 name = "astroid"
-version = "2.6.6"
+version = "2.8.2"
 description = "An abstract syntax tree for Python with inference support."
 category = "dev"
 optional = false
@@ -24,6 +24,7 @@ python-versions = "~=3.6"
 
 [package.dependencies]
 lazy-object-proxy = ">=1.4.0"
+typing-extensions = {version = ">=3.10", markers = "python_version < \"3.10\""}
 wrapt = ">=1.11,<1.13"
 
 [[package]]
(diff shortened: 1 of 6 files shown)


shepherdjay:
> Awesome getting to this fix fast. 👍  Followup and get that linting fixed on that test and we're good to go!

Thanks - I made an attempt to fix the linting on the go. If it still doesn't pass will pass it through auto king end of today when I get back to my development station.

PR #486, opened 2021-10-26
Refactor Types using CapWords naming style

### LINKED ISSUE #482: 🐞 Types cased like Constants

### Current Behavior

There are some types such as `GUARDIAN_ID` that are cased incorrectly. These variables function as types and not constants and therefore should be using "CapWords" since these are actually functioning as Type's. 

https://www.python.org/dev/peps/pep-0008/

### Expected Behavior

Global Constants should be capitalized with underscores
`THRESHOLD = 3` 

Classes and Types should be like this:
`GuardianId = str` 

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: Any
```

### Anything else?

It's probably helpful just to look at the types from this level for speed. 

https://github.com/microsoft/electionguard-python/blob/main/src/electionguard/__init__.py
https://github.com/microsoft/electionguard-python/blob/main/src/electionguard_tools/__init__.py
https://github.com/microsoft/electionguard-python/blob/main/src/electionguard_verify/__init__.py

faraazb:

### Issue
Fixes #482 

### Description
Some Types were in the UPPER_CASE_WITH_UNDERSCORES naming style like Constants. Such Types were refactored to use CapWords naming style.

### Testing
Only consists of refactoring; I performed all the included tests, and got a DeprecationWarning which I don't think is related.
```
====================================== warnings summary ======================================
tests/unit/test_ballot_compact.py::TestCompactBallot::test_compact_submitted_ballot
  <containing_directory>\electionguard-python\src\electionguard\ballot.py:852: DeprecationWarning: NotImplemented should not be used in a boolean context
    isinstance(other, SubmittedBallot)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
=================== 187 passed, 1 skipped, 1 warning in 301.61s (0:05:01) ====================
```


faraazb:
Thank you! @keithrfung @SteveMaier-IRT

PR #487, opened 2021-10-27
Add bounds to BaseElement class

### LINKED ISSUE #463: 🧮  Revisit BaseElement, ElementModQ, ElementModP

### Suggestion

`ElementModQ` and `ElementModP` sit at the base of many of numbers and serve as a bounded integer that most of the math uses. There have been concerns raised around the inheritance from an `int`. 

A large part of this concern is the possibility for a developer to use `ElementModQ` or `ElementModP` in typical `int` math functions and exceed the lower or higher bound without knowing it. The math functions supplied in the library won't run into this issue, but since the `+` or `*` operator can be used on an `int`, it could arise. 

For example: A developer could harmlessly do:
```py
a = ElementModQ(MAX_Q)
b = 1
c = a + b # should exceed max
```
Instead of using the math functions we supply such as `add_q`. 

### Possible Implementation

There are two possible solutions, both will require a significant amount of tests.

**1. Override existing math functions**
Ensure any typical math functions can't be used so that when a developer is using `ElementModP` and `ElementModQ` it will catch any numbers that exceed the bounds. 

The most concerning ones that are concerning are add, mult, subt, and pow. 
https://docs.python.org/3/library/math.html
https://diveintopython3.net/special-method-names.html

**2. Turn BaseElement into a Number class**
This can mean making an abstract Number class or simply just providing all the methods in BaseElement. 
Essentially this means making a custom number class that serializes like an `int`. 
https://www.python.org/dev/peps/pep-3141/

It may be wise to at this juncture to also implement `__ceil__` and `__floor__` instead of lower and upper bound. 

#### Math

This case requires significant testing. Here are a list of suggestions. 

Breaking the ceiling:
```py
should_break = ElementModQ(MAX_Q) + 1 
should_also_break = 1 + ElementModQ(MAX_Q)
```

Breaking the floor:
```py
should_break = ElementModQ(0) - 1 
should_also_break = 0 - ElementModQ(1)
```

These should be thorough and test different variations of ElementModP and ElementModQ while interacting with `int` and other common math functions. 

#### Serialization

We want to ensure there is no serialization code in the base `electionguard` repository. This is to give flexibility to other developers who may choose to use our library differently. 

Currently our `electionguard-api-python` uses FastAPI thus pydantic. Pydantic like other json serialization tools uses `asdict` to ensure that an object can be serialized as plain Javascript object. Similarly, we have many other developers that use pickle with our library. We must support both use cases.

As a simple example, using `pickle` OR `asdict` with `json.dumps` on
```py
class Example:
      value: ElementModQ
```
should serialize to and deserialize from
```json
{
    "value": 123
}
```

ronnie-llamado:

### Issue
Fixes #463

### Description
Adds bounds to BaseElement. An `OverflowError` will be thrown if the bounds are exceeded using the mathematical dunder methods ( `__add__`, `__mul__`, etc).

#### To Do
- [x] Add tests using standard operators (+, -, *, **)

### Testing
`make coverage`


tests/property/test_group.py
```diff
@@ -149,8 +149,8 @@ def test_in_bounds_q(self, q: ElementModQ):
     @given(elements_mod_p())
     def test_in_bounds_p(self, p: ElementModP):
```
keithrfung, line 150: I actually think this should be creating integers not ElementModP's and Q's with the way we are using this.

src/electionguard_tools/helpers/serialize.py
```diff
@@ -122,6 +123,8 @@ def custom_encoder(obj: Any) -> Any:
         new_dict = asdict(obj)
         obj = _recursive_replace(new_dict, int, _get_int_encoder())
         return obj
+    elif isinstance(obj, BaseElement):
+        return obj.to_hex()
```
keithrfung, line 127: This is the piece that I was trying to avoid. Having to check for `BaseElement` to do the serialization. This is essentially what we had prior to the switch to `int`.

src/electionguard/election_polynomial.py
```diff
@@ -81,11 +81,11 @@ def compute_polynomial_coordinate(
     :return: Polynomial used to share election keys
     """
 
-    exponent_modifier = ElementModQ(exponent_modifier)
+    exponent_modifier_q = ElementModQ(exponent_modifier)
```
keithrfung, line 84: This is a good refactor, but should this be in this PR? Did this break anything?

pyproject.toml
```diff
@@ -92,6 +92,11 @@ disable = '''
 branch = true
 source = ["src/electionguard"]
 
+[tool.coverage.report]
```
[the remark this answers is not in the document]
keithrfung: I don't this is the best way to go about this. Explain to me more how this solved the problem.

keithrfung (changes requested):
See the change list. This is a good first move. We can discuss this further if you need it.

src/electionguard/group.py
```diff
@@ -17,17 +20,133 @@
...
+
+    def __add__(self, other: Any) -> BaseElement:
+        """a + b"""
+        if isinstance(other, (self.__class__, int)):
+            return self.__class__(int(self) + int(other))
+
+        return NotImplemented
+
+    def __radd__(self, other: Any) -> BaseElement:
+        """a + b"""
```
SteveMaier-IRT, line 58: Should probably document this as b+a since you are adding then in the other order in the method.  This happens with other reverse functions below as well.  Minor issue.

keithrfung (commented):
I've taking a look at this. First things first, this is an awesome PR. There are some things that I really like such as the thoroughness of the tests and methods on the `BaseElement`.

I think this really shows that the solution isn't as simple as moving to the `Number` abstract class. The issue is really designed to help deal with the `PowRadix` situation we are moving towards AND also resolve mathematical operations surrounding `BaseElement`. This seems to emphasize at least as the serialization currently works we should focus on the mathematical operations. 

Here's a quick example I cobbled together:
Screen Shot 2021-11-09 at 2 49 22 PM


```py
from dataclasses import asdict, dataclass
from numbers import Number
# =============INTEGER========================
class IntElement(int):
  def __new__(cls, value: int):
    return super(IntElement, cls).__new__(cls, value)

class IntPowRadix(IntElement):
  _pow_radix: int = 0

  def __new__(cls, value: int):
    self =  super(IntPowRadix, cls).__new__(cls, value)
    self._pow_radix = 12345 # Imagine calculation here
    return self

# =============NUMBER========================
@dataclass
class NumberElement(Number):

  value: int = 0

  def __new__(cls, value: Number):
    self = super(NumberElement, cls).__new__(cls)
    self.value = value
    return self

@dataclass
class NumberPowRadix(NumberElement):

  value: int = 0
  _pow_radix: int = 0

  def __new__(cls, value: Number):
    self = super(NumberPowRadix, cls).__new__(cls, value)
    self.value = value
    self._pow_radix = 12345 # Imagine calculation here
    return self

@dataclass
class Wrapper:
  integer: IntElement

  integer_pow_radix: IntPowRadix
  number: NumberElement
  number_pow_radix: NumberPowRadix

# Arrange
test_int = IntElement(123)
test_number = NumberElement(456)
test_integer_pow_radix = IntPowRadix(789)
test_number_pow_radix = NumberPowRadix(789)
test_wrapper = Wrapper(test_int, test_integer_pow_radix, test_number, test_number_pow_radix)

#Act
test_print = asdict(test_wrapper)
print(test_print)
```

tests/property/test_group.py
```diff
@@ -143,8 +143,8 @@ def test_simple_powers(self):
     @given(elements_mod_q())
     def test_in_bounds_q(self, q: ElementModQ):
         self.assertTrue(q.is_in_bounds())
-        too_big = q + get_small_prime()
-        too_small = q - get_small_prime()
+        too_big = int(q) + get_small_prime()
+        too_small = int(q) - get_small_prime()
```
keithrfung, line 147: Why does this have to be cast to an int? 

My thought would be 

Assert that:
- `ElementModQ + small prime int => too big ElementModQ` 
     - If ElementModQ is too big after addition, Overflow error should be tossed

- `int + small prime int = > too big int`
    - If ElementModQ is constructed with too big int, Overflow error should be tossed
    - Also validate the is_in_bounds works

tests/property/test_group.py
```diff
@@ -157,8 +157,8 @@ def test_in_bounds_q(self, q: ElementModQ):
     @given(elements_mod_p())
     def test_in_bounds_p(self, p: ElementModP):
         self.assertTrue(p.is_in_bounds())
-        too_big = p + get_large_prime()
-        too_small = p - get_large_prime()
+        too_big = int(p) + get_large_prime()
+        too_small = int(p) - get_large_prime()
```
keithrfung, line 161: Similar question to above.

Why does this have to be cast to int?

src/electionguard/group.py
```diff
@@ -17,7 +19,57 @@
...
+            return value
+
+        @wraps(obj)
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            return convert(obj(*args, **kwargs))
+
+        return wrapper
+
+
+class BaseElement(int, metaclass=BaseElementMeta):
```
keithrfung, line 72: We've hit `metaclass`. Oh noes. This isn't quite what I was thinking. Let's discuss. Reach out view email.

keithrfung (changes requested):

keithrfung:
Closing this. We will revisit in call.

PR #492, opened 2021-11-09
Fix (and add tests for) BaseElement.__ne__

### LINKED ISSUE #490: 🐞  Double negative on not equal function on BaseElement

### Current Behavior

https://github.com/microsoft/electionguard-python/blob/6bf466f3cf8b3f47cd29e1c549f86ad1758b6221/src/electionguard/group.py#L32-L38

### Expected Behavior

The not equal (`__ne__`) function should correctly handle not equal. 

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: Any
```

### Anything else?

Please add a test in `test_group.py` to ensure this bug does not reappear.

ronnie-llamado:

### Issue
Fixes #490 

### Description
Fix double negative inside of `BaseElement.__ne__`

### Testing
`make coverage`

### DIFF
--- a/src/electionguard/group.py
+++ b/src/electionguard/group.py
@@ -31,7 +31,7 @@ class BaseElement(ABC, int):
 
     def __ne__(self, other: Any) -> bool:
         """Overload != (not equal to) operator."""
-        return isinstance(other, (BaseElement, int)) and not int(self) != other
+        return not self == other
 
     def __eq__(self, other: Any) -> bool:
         """Overload == (equal to) operator."""
--- a/tests/property/test_group.py
+++ b/tests/property/test_group.py
@@ -47,12 +47,16 @@ class TestEquality(BaseTestCase):
 
     @given(elements_mod_q(), elements_mod_q())
     def test_p_not_equal_to_q(self, q: ElementModQ, q2: ElementModQ):
+        i = int(q)
+        i2 = int(q2)
         p = ElementModP(q)
         p2 = ElementModP(q2)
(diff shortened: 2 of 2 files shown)


src/electionguard/group.py
```diff
@@ -31,7 +31,7 @@ def __new__(cls, elem: Union[int, str], check_within_bounds: bool = True):  # ty
 
     def __ne__(self, other: Any) -> bool:
         """Overload != (not equal to) operator."""
-        return isinstance(other, (BaseElement, int)) and not int(self) != other
+        return not self == other
```
keithrfung, line 34: How does this react to items that aren't `int` or `BaseElement`? 

For example:
string = str(q)
self.assertEqual(q, string)

ronnie-llamado: Since strings will return `False` in `__eq__`, any comparison against strings will not be equal in this setup.

Here's the StackOverflow inspiration: https://stackoverflow.com/questions/4352244/should-ne-be-implemented-as-the-negation-of-eq

Should the example you provided be True, or was that just for example purposes? `ElementModQ(1) == "1"`?

keithrfung: That was for example purposes to make sure we can remove the `isinstance` call.

keithrfung (approved):
Perfect example of me after my booster shot giving approval.

@SteveMaier-IRT Can I get a second approval?

SteveMaier-IRT (approved):

PR #496, opened 2021-11-16
Dataclass incompatible: Fix fields in manifest contactInfo

### LINKED ISSUE #495: Generate correct email field of ContactInfo for property test

### Current Behavior

When we use property test to generate fake manifest contact info, we directly uses emails() method from hypothesis.strategies package to create a email format String. However, the expected data type of emails in ContactInformation is List[AnnotatedString]. 

This leads to crush when we try to deserialize a manifest json file to the python Manifest object. 

### Expected Behavior

For each email in contact info, the property test should generate a list of annotated string that contains both annotation and value. 

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: MacOS 10.14.6
```

### Anything else?

_No response_

Xin128:
The initial manifest creation draws a String using emails() methods from hypothesis strategies, but it requires a list of annotated string. I added a annotated_emails method to draw a two_letter_code for annotation and a email-format string for value. 

### Issue

Fixes #495 

### Description
Generate a list of annotated string for email field in contact info. 
### Testing
make test

### DIFF
--- a/src/electionguard_tools/strategies/election.py
+++ b/src/electionguard_tools/strategies/election.py
@@ -174,6 +174,14 @@ def reporting_unit_types(draw: _DrawType):
     n = draw(sampled_from(ReportingUnitType))
     return ReportingUnitType(n)
 
+@composite
+def annotated_emails(draw: _DrawType):
+    """
+       Generates a `Email` object with an arbitrary two-letter string as annotation and an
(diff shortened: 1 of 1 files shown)


keithrfung (approved):
Solid improvement. Thanks for the fix!

PR #504, opened 2021-12-03
Fix equality of InternalManifests constructed from same Manifest

### LINKED ISSUE #501: 🐞  InternalManifests constructed from same Manifest not equal

### Current Behavior

[@PradyumnaKrishna mentioned in his PR](https://github.com/microsoft/electionguard-python/pull/497/files#r756633619) after using the `InternalManifest` constructor to build from a `Manifest` that the equality seems to be broken. This was utilizing the hypothesis strategy seen below. 

https://github.com/microsoft/electionguard-python/blob/4b275cf4aeb12b17f6526806403b2830ff6db600/src/electionguard_tools/strategies/election.py#L693-L722

This could be an issue with the equality operator, the InternalManifest constructor or perhaps something else. This issue is to investigate and if repeated, resolve. 

### Expected Behavior

`InternalManifest` that are constructed using same `Manifest` should be equal. 

### Steps To Reproduce

```py
@given(integers(1, 3).flatmap(lambda n: elections_and_ballots(n)))
    def test_manifest_equality(self, election_details: ElectionsAndBallotsTupleType):
        # Arrange
        (
            manifest,
            internal_manifest,
            _ballots,
            _secret_key,
            _context,
        ) = election_details
        new_internal_manifest = InternalManifest(manifest)
        self.assertEqual(internal_manifest, new_internal_manifest)
```

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

PradyumnaKrishna:
### Issue

Fixes #501 

### Description
The cause of inequality of InternalManifests constructed from same Manifest was
https://github.com/microsoft/electionguard-python/blob/db1287d66c8aabf70a9d338bf6c249578906b3df/src/electionguard_tools/strategies/election.py#L630

Since lists are call by reference in python, shuffling `ballot_selections` affects `contest`. This changes the hash of `Contests` and creates a difference in `manifest_hash` for the `InternalManifest`.

Creating a copy of `ballot_selections` fixes this issue and made improvements to the test that uses this method.

### DIFF
--- a/tests/property/test_verify.py
+++ b/tests/property/test_verify.py
@@ -16,7 +16,6 @@ from electionguard.elgamal import ElGamalKeyPair
 from electionguard.encrypt import EncryptionMediator, encrypt_ballot
 from electionguard.key_ceremony import CeremonyDetails
 from electionguard.key_ceremony_mediator import KeyCeremonyMediator
-from electionguard.manifest import InternalManifest
 from electionguard.tally import tally_ballots
 from electionguard.type import GuardianId
 from electionguard.utils import get_optional
@@ -121,7 +120,7 @@ class TestVerify(BaseTestCase):
         self.assertTrue(verification.verified)
 
     @settings(
-        deadline=timedelta(milliseconds=2000),
+        deadline=timedelta(milliseconds=10000),
         suppress_health_check=[HealthCheck.too_slow],
         max_examples=10,
         # disabling the "shrink" phase, because it runs very slowly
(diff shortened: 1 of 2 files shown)


tests/property/test_verify.py
```diff
@@ -121,7 +120,7 @@ def test_verify_decryption(self):
         self.assertTrue(verification.verified)
 
     @settings(
-        deadline=timedelta(milliseconds=2000),
+        deadline=timedelta(milliseconds=10000),
```
keithrfung, line 123: This is a significant bump of time. Does this need to be 5 times larger?

PradyumnaKrishna: I am not sure about time, saw 10000 ms on other test using same hypothesis.
https://github.com/microsoft/electionguard-python/blob/37b55504b62c07cd6083ddcbebb8c6af84e574e6/tests/property/test_tally.py#L39
and saw a test failed once on the action because it exceeds the deadline of 2000.00ms

keithrfung (approved):

PR #511, opened 2021-12-15
Allow specifying sample to use when running the generator

### LINKED ISSUE #200: Allow generation of sample data with manifests
Modify `make generate-sample-data` to take in a manifest so we can very quickly create multiple sample data sets that are published per release instead of creating them manually and having in repository. 

This will likely be two changes. 
1. The `ElectionSampleDataGenerator` needs to be modified
2. The make command needs to be modified

AddressXception:

### Issue
*Link your PR to an issue*

Fixes #200

### Description
Add CLI params to the sample generation process to define a spec version and the sample set to use. when setting up the environment, pull down the sample data from the existing spec release from the `electionguard` spec repo.

This PR satisfies the requirement of making the sample generator configurable, however due to ongoing serialization changes, only the `hamilton-general` sample can actually be run at this time.

### Testing
1. `make environment`
2. `make generate-sample-data`

Expect: sample data is output correctly to the `/results` folder.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,10 +19,10 @@ environment:
 	poetry config virtualenvs.in-project true 
 	poetry install
 	@echo 🚨 Be sure to add poetry to PATH
-	ifeq ($(OS), Windows)
-		choco install wget
-		choco install unzip
-	endif
+ifeq ($(OS), Windows)
+	choco install wget
+	choco install unzip
+endif
 	wget https://github.com/microsoft/electionguard/releases/download/v0.95.0/sample-data.zip
 	unzip -o sample-data.zip
 	unzip sample-data.zip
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -19,6 +19,8 @@ environment:
 	poetry config virtualenvs.in-project true 
 	poetry install
 	@echo 🚨 Be sure to add poetry to PATH
+	wget -O sample-data.zip https://github.com/microsoft/electionguard/releases/download/v0.95.0/sample-data.zip
+	unzip sample-data.zip
```
SteveMaier-IRT, line 23: you might want to add a -o to this to allow the files to overwrite a previous download.  This is what is happening on the cpp project as well.

SteveMaier-IRT (changes requested):
Adding parameters for the unzip to overwrite (-o) should fix the build and would be the only thing to change here.

SteveMaier-IRT (approved):

PR #530, opened 2022-02-07
Feature/remove auxiliary keys

### LINKED ISSUE #507: 🔥. Remove Auxiliary Keys

### Suggestion

`Auxiliary` keys and the key pair were originally used to secure messaging between the guardians. For ElectionGuard 1.0.0, it has been decided that these are not necessary since the information often secured is already made public. 

### Possible Implementation

- [ ] Remove Auxiliary Keys (https://github.com/microsoft/electionguard-python/blob/main/src/electionguard/auxiliary.py)
- [ ] Update any documentation within this repository reference auxiliary keys
- [ ] Remove any references and update methods that use auxiliary keys

### Anything else?

_No response_

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #507 

### Description
Auxiliary Keys are no longer relevant for ElectionGuard SDK 1.0. This request removes the auxiliary keys from encrypt and decrypt functions. Subsequently this hits the key ceremony and guardians. 

The RSA and identity encrypt are also removed at the same time. These were only created to support the auxiliary keypairs. This reduces code maintenance and code size.

### Testing
- Existing Test Suite should be sufficient. See CI


src/electionguard/key_ceremony_mediator.py
```diff
@@ -225,15 +195,6 @@ def _receive_election_public_key(self, public_key: ElectionPublicKey) -> None:
         """
         self._election_public_keys[public_key.owner_id] = public_key
 
-    def _all_election_public_keys_available(self) -> bool:
```
SteveMaier-IRT, line 228: Is the removal of this function needed since it is not using aux keys.  Just not sure if there are needs for this.

keithrfung: Originally there was
`all_public_keys_received` which relied on `_all_election_public_keys_available` and `_all_auxiliary_public_keys_available`

Now it's just `all_guardian_keys_received` since we only need one keyset.

PR #537, opened 2022-02-08
✅ Ensure all python commands reference Python3

### LINKED ISSUE #515: 🐞  For MacOS users, prefer python3 to python in MakeFile

### Current Behavior

On MacOS, the install and build scripts use `python` command which for MacOS without configuration is Python 2.7.

### Expected Behavior

The scripts should reference Python3 which can be done by using `python3` uniformly.

### Environment

```markdown
- OS: MacOS 11.6.2
```

### Resolution

Replace all `python` commands with `python3` in MakeFile

https://github.com/microsoft/electionguard-python/blob/a4d6cb43cb1573e19c52d4560d06fa38ecfe59ee/Makefile#L4
https://github.com/microsoft/electionguard-python/blob/a4d6cb43cb1573e19c52d4560d06fa38ecfe59ee/Makefile#L6
https://github.com/microsoft/electionguard-python/blob/a4d6cb43cb1573e19c52d4560d06fa38ecfe59ee/Makefile#L8
https://github.com/microsoft/electionguard-python/blob/a4d6cb43cb1573e19c52d4560d06fa38ecfe59ee/Makefile#L17
https://github.com/microsoft/electionguard-python/blob/a4d6cb43cb1573e19c52d4560d06fa38ecfe59ee/Makefile#L114

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #515 

### Description
Change all commands to reference Python 3 specifically. This is to simplify issues in MacOS where the user has to specify in their environment that `python` refers to Python 3 instead of the default Python 2. 

### Testing
Run any of the commands. This should happen by default by the CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,11 @@
 .PHONY: all environment openssl-fix install install-gmp install-gmp-mac install-gmp-linux install-gmp-windows install-mkdocs auto-lint validate test test-example bench coverage coverage-html coverage-xml coverage-erase generate-sample-data
 
 CODE_COVERAGE ?= 90
-OS ?= $(shell python -c 'import platform; print(platform.system())')
+OS ?= $(shell python3 -c 'import platform; print(platform.system())')
 ifeq ($(OS), Linux)
-PKG_MGR ?= $(shell python -c 'import subprocess as sub; print(next(filter(None, (sub.getstatusoutput(f"command -v {pm}")[0] == 0 and pm for pm in ["apt-get", "pacman"])), "undefined"))')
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@ all: environment install build validate auto-lint coverage
 environment:
 	@echo 🔧 ENVIRONMENT SETUP
 	make install-gmp
-	python -m pip install -U pip
+	python3 -m pip install -U pip
```
SteveMaier-IRT, line 17: Verified that the pip3 command is not needed but we might need some instructions on it in the setup.

SteveMaier-IRT (approved):

PR #541, opened 2022-02-14
♻️  Add GuardianId to LagrangeCoefficientRecord

### LINKED ISSUE #520: ✨The Guardian ids should be added to the coefficients JSON file

### Suggestion

Currently the lagrange coefficients are simply serialized as a list:
```
{
    "coefficients": [
        "05",
        "FFE7",
        "0A",
        "FFEC",
        "01"
    ]
}
```

Which Guardian do they reference? One could assume that one can read the GuardianRecords and sort them (on sequence_order?) and match them up.

However, what happens when the number of decrypting guardians is not the full number? Should only the "available" GuardianRecords be saved into the election record? This is not described anywhere, AFAIK. And the library itself does not enforce any such usage. One could imagine that the GuardianRecords might be conveniently stored during the KeyCeremony. Only during decryption do you know which are available for decrypting. Only when you know the set of  available Guardians can you compute the lagrange coefficients .

Seems more robust to store the guardian ids with the lagrange coordinates. This would also unambiguously specify which guardians were part of the decryption.

### Possible Implementation
```

@dataclass
class LagrangeCoefficientsRecord:
    coefficients: List[ElementModQ]
    guardian_ids: List[str]
```

### Anything else?

Not sure if there are any python tests for the case when all guardians are not used for decryption. TestEndToEndElection seems to only have the full case? Would be good to generate a canonical test election record for both full and not full case, so verifiers can test their code.

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #520 

### Description
Change LagrangeCoefficients to stay a dictionary on output. This actually reduces some code complexity and quickly allows the guardian id to exist for the LagrangeCoefficients.

Sample data will have to regenerated following this. An issue will be made in `electionguard`

### Testing
The integration test is the best place for this to be caught.

### DIFF
--- a/src/electionguard/election_polynomial.py
+++ b/src/electionguard/election_polynomial.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import List
+from typing import Dict, List
 
 from .elgamal import ElGamalKeyPair
 from .group import (
@@ -17,6 +17,7 @@ from .group import (
     ZERO_MOD_Q,
 )
 from .schnorr import make_schnorr_proof, SchnorrProof
+from .type import GuardianId
 
 SecretCoefficient = ElementModQ  # Secret coefficient of election polynomial
 PublicCommitment = ElementModP  # Public commitment of election polynomial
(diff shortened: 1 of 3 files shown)


keithrfung:
@JohnLCaron This is a slight modification on what you asked for but would love your thoughts. This is a short term solution but I think feasible in all languages and quick.

SteveMaier-IRT (approved):

JohnLCaron:
Simple and elegant. Seems fine to me.

PR #543, opened 2022-02-15
⚙️ Add Configuration to Context

### LINKED ISSUE #386: ✨ Election Configuration of Edge Cases
## Feature Request

**Description**
A system for handling allowing or disallowing of certain edge cases.
Examples includes:
- null votes
- undervotes
- overvotes

**Expected Solution**
Currently, the only concern is handling overvotes. A `Config` will be added to the `Context` allowing overvotes to be enabled. This can then be part of setting up the election

For example:
```
class Config():
      allow_over_votes: bool
```

The `Config` itself should have a constructor that defaults to current supported setup.
```py
allow_over_votes: False
```

If another option is selected other than these, they should throw a NotImplemented with a message indicating that the feature is not available, until the feature is implemented.

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #386 

### Description
Add configuration to allow_overvotes to be set

### DIFF
--- a/src/electionguard/election.py
+++ b/src/electionguard/election.py
@@ -1,6 +1,6 @@
 """Context for election encryption."""
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Dict, Optional
 
 from .constants import get_small_prime, get_large_prime, get_generator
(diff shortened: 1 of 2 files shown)


src/electionguard/election.py
```diff
@@ -55,6 +66,9 @@ class CiphertextElectionContext:
     extended_data: Optional[Dict[str, str]]
     """Data to allow extending the context for special cases."""
 
+    configuration: Configuration = field(default_factory=Configuration)
```
SteveMaier-IRT, line 69: Will we need to change any of our sample data or tests with this new field?

SteveMaier-IRT (approved):
Good start for the configuration.  Will be easy to add more fields when we need them.

src/electionguard/election.py
keithrfung: Yes, we will need to change all the sample data. We can do this post 1.4 being release though instead of chasing our tails.

Issue #551, opened 2022-02-25
🔐  Hashed ElGamal


keithrfung:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Suggestion

In the system to support secret sharing and encrypting extended data, there was originally auxiliary keys that used rsa to do encryption. Hashed ElGamal is the replacement that doesn't an additional key but allows reusing the guardian key for secret sharing and the existing encryption key for encrypting extended data. This Hashed ElGamal will use a setup involving HMAC if you are interested in researching more. 

### Possible Implementation

This will be a really high overview. There should be a class `HashedElGamalCiphertext` that will inherit from the existing `ElGamalCiphertext`. There will be some code to generate hmac keys. The Ciphertext will include methods for decryption in the same way the existing protocol for votes exists.

Issue #385, opened 2021-07-15
✨  Ensure Overvotes can be passed into system


keithrfung:
## Feature Request

**Description**
Overvotes are when there are more votes selected than are allowed. For example, where only one can be selected there are 3 selected. The ballot once it enters still needs to be able to be challenged by a voter to show the system is verifiable. Currently, an overvote will fail ballot validity.

**Proposed Solution**

Steps:
The current encryption code ensures overvotes are not allowed. Under certain configurations (See #386), allow overvotes to pass through encryption.

If overvotes are allowed:
1. Placeholders are recorded for all the votes. Example: selection limit = 1, 1 placeholder; selection limit = 2, 2 placeholders.
2. Contest is marked as overvote using `encrypted_data` and holds information about contests selected.
3. Previous votes are set to 0

**EXAMPLE:**
**Overvoted Contest (Pick 2)**
extended_data: null
selections:
- Candidate A - X
- Candidate B - X
- Candidate C - X
- Placeholder - O
- Placeholder - O

**Resolved Contest**
extended_data: $overvote_info
selections:
  - Candidate A - O
  - Candidate B - O
  - Candidate C - O
  - Placeholder - X
  - Placeholder - X

## Documentation 

From Josh Benaloh (@benaloh)

In some use cases, it is desirable to capture details about overvotes without affecting tallies.  In ElectionGuard, an overvote in a contest (i.e., the number of selections exceeds the selection limit) causes the following things to happen.

1. All selectable options in that contest are set to zero.
2. All placeholder options in that contest are set to one.
4. The extended data field for the overvoted contest contains the following data:
      a. The first byte is set to x01 as a tag to indicate an overvote.
      b. The remaining bytes form a bit array indicating the indices of selections made by the voter.
5. If the voter selected one or more write-ins among the overvotes, any such write-in text is retained in the text field of the associated write-in even though the associated write-in selection is set to zero.

The lengths of the extended data field(s) and the write-in text fields are set in the ballot coding file.  To accommodate preservation of overvote data as described here, the bit-length of each extended data field must be at least 8 greater than the number of selectable options in that contest (including write-ins).

As will be described in greater detail elsewhere, the encryption of all non-binary data on a ballot, (extended data fields, write-in text fields, etc.) is done as follows.

[...]

PR #556, opened 2022-03-01
🔒 Add extended data field to contest

### LINKED ISSUE #553: ✨ Add extended_data field to Contest

### Suggestion

The contests on a ballot need an extended_data field to store information particularly for the overvotes use case. This data as detailed in #385 can be used to store information that the contest was overvoted. This is only to add the fields and segment some of the work out and should be accessible for coders starting out with ElectionGuard.

### Possible Implementation

The existing implementation from the ballot selections should be able to be replicated for contest

Example from `PlaintextBallotSelection`
https://github.com/microsoft/electionguard-python/blob/c59fddcd89f85619ba801d6c38b1ea7f08210951/src/electionguard/ballot.py#L74-L77

Example from `CiphertextBallotSelection`
https://github.com/microsoft/electionguard-python/blob/c59fddcd89f85619ba801d6c38b1ea7f08210951/src/electionguard/ballot.py#L176-L177

These should be added to the similarly named
`PlaintextBallotContest` and `CiphertextBallotContest`

### Anything else?

_No response_

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #553 

### Description
Add extended data field to contest. This should later be updated to encrypt and decrypt at a later point. Similar non working field to `extended_data` on selection until issues #385, and #551 are complete

### Testing
Existing test suite should suffice.

### DIFF
--- a/src/electionguard/ballot.py
+++ b/src/electionguard/ballot.py
@@ -70,7 +70,6 @@ class PlaintextBallotSelection(OrderedObjectBase):
     is_placeholder_selection: bool = field(default=False)
     """Determines if this is a placeholder selection"""
 
-    # TODO: ISSUE #35: encrypt/decrypt
     extended_data: Optional[ExtendedData] = field(default=None)
     """
     an optional field of arbitrary data, such as the value of a write-in candidate
@@ -172,7 +171,6 @@ class CiphertextBallotSelection(
     proof: Optional[DisjunctiveChaumPedersenProof] = field(default=None)
     """The proof that demonstrates the selection is an encryption of 0 or 1, and was encrypted using the `nonce`"""
 
-    # TODO: ISSUE #35: encrypt/decrypt
     extended_data: Optional[ElGamalCiphertext] = field(default=None)
     """encrypted representation of the extended_data field"""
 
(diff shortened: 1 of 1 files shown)


src/electionguard/ballot.py
```diff
@@ -172,7 +171,6 @@ class CiphertextBallotSelection(
     proof: Optional[DisjunctiveChaumPedersenProof] = field(default=None)
     """The proof that demonstrates the selection is an encryption of 0 or 1, and was encrypted using the `nonce`"""
 
-    # TODO: ISSUE #35: encrypt/decrypt
```
keithrfung, line 175: We're trying to remove these and this is now captured in a new issue.

src/electionguard/ballot.py
```diff
@@ -172,7 +171,6 @@ class CiphertextBallotSelection(
...
     extended_data: Optional[ElGamalCiphertext] = field(default=None)
```
SteveMaier-IRT, line 174: Is a different issue for changing this to HashedElgamalCiphertext instead of the ElGamalCiphertext?

SteveMaier-IRT (changes requested):
Only a single comment around the data type for the encrypted.

PR #557, opened 2022-03-04
Feature/hash elgamal

### LINKED ISSUE #551: 🔐  Hashed ElGamal

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #551 

### Description
This pull request implements just the Hashed Elgamal encryption and decryption. This requires hmac to be used. To avoid the direct reference in the elgamal file. This is shifted to it's own file. 

Michael gave me a hand with some of the code but this still needs a bit more work. 

**Note:** The type testing is currently failing. I didn't have quite enough time to polish this out. 

### Testing
There are two new tests added for this. 
1. `test_elgamal.py`
2. `test_hmac.py`


src/electionguard/elgamal.py
```diff
@@ -154,6 +206,71 @@ def elgamal_encrypt(
     return ElGamalCiphertext(pad, data)
 
 
+_MAC_KEY_SIZE = 256
```
keithrfung, line 209: These should likely get moved to top of file.

src/electionguard/elgamal.py
```diff
@@ -154,6 +206,71 @@ def elgamal_encrypt(
...
+    log_info(f": data: {data}")
+    log_info(f": mac: {mac}")
+
+    return HashedElGamalCiphertext(pad, data, mac)
+
+
+_BLOCK_SIZE = 32
+
+
+def _get_chunks(message: bytes):
```
keithrfung, line 260: This chunking function is to reduce some code complexity. It's not the prettiest for sure. Probably could use some work.

src/electionguard/elgamal.py
SteveMaier-IRT: moved to top.

src/electionguard/elgamal.py
SteveMaier-IRT: moved.

src/electionguard/elgamal.py
```diff
@@ -128,12 +186,12 @@ def elgamal_combine_public_keys(keys: Iterable[ElementModP]) -> ElementModP:
 
 
 def elgamal_encrypt(
-    m: int, nonce: ElementModQ, public_key: ElementModP
+    message: int, nonce: ElementModQ, public_key: ElGamalPublicKey
```
lprichar, line 189: ❤️

lprichar (approved):

PR #560, opened 2022-03-30
🧮  Precompute Discrete Log Cache with exponent

### LINKED ISSUE #559: ✨ Precompute Discrete Log Cache

### Suggestion

A more direct approach on creating a discrete log. Use the exponent or the number of ballots to generate the table as opposed to using the value as the parameter to start the calculation.

Add a `max_votes` configuration for the max number of votes or in our case ballots that are being calculated instead of relying on the `DLOG_MAX` to allow for smaller configurations and customized numbers. 

Create methods to calculate cache directly up to an exponent to help preload a cache. Should be both async and not. 

### Possible Implementation

_No response_

### Anything else?

_No response_

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #559 

### Description
- Add a precompute for discrete log cache for exponent instead of relying on upward bound lazy evaluation. 
- Add a setting for max_votes
- Added methods to allow avoiding the cache if necessary.

### Testing
- test_discrete_log.py
- Integration tests


SteveMaier-IRT (approved):
looks good.

src/electionguard/discrete_log.py
```diff
@@ -58,29 +74,64 @@ async def discrete_log_async(
         if element in cache:
             return (cache[element], cache)
 
-        _cache = compute_discrete_log_cache(element, cache)
+        _cache = compute_discrete_log_cache(element, cache, max_exponent)
         return (_cache[element], _cache)
 
 
+def precompute_discrete_log_cache(
```
lprichar, line 81: There's a lot of new functionality here, should this have a covering unit test?

lprichar (approved):
lgtm

src/electionguard/discrete_log.py
keithrfung: Unit tests added

tests/property/test_discrete_log.py
```diff
@@ -37,7 +38,7 @@ class TestDiscreteLogFunctions(BaseTestCase):
     """Discrete log tests"""
 
     @given(integers(0, 100))
-    def test_uncached(self, exp: int):
+    def test_uncached(self, exp: int) -> None:
```
keithrfung, line 41: This is boyscouting. The tests should all be typed.

tests/property/test_discrete_log.py
```diff
@@ -80,20 +81,50 @@ def test_cached_one_async(self):
...
+        minimum_cache_size = exponent + 1
+        element = g_pow_p(exponent)
+
+        # Act
+        DiscreteLog().set_lazy_evaluation(False)
+        DiscreteLog().precompute_cache(exponent)
+        calculated_exponent = DiscreteLog().discrete_log(element)
+
+        # Assert
+        self.assertGreaterEqual(len(DiscreteLog().get_cache()), minimum_cache_size)
```
keithrfung, line 123: This is related to the singleton. The test backtracks and does 998 and then repeats one so greater than or equal to is necessary. This is okay though because we only want to ensure the cache is created correctly and covers this case.

src/electionguard/discrete_log.py
```diff
@@ -6,16 +6,39 @@
...
-"""The max number to calculate.  This value is used to stop a race condition."""
+DiscreteLogCache = Dict[ElementModP, int]
+
+_DLOG_MAX_EXPONENT = 100_000_000
+"""The max exponent to calculate.  This value is used to stop a race condition."""
+
+_INITIAL_CACHE = {ONE_MOD_P: 0}
+
+
+class DiscreteLogExponentError(ValueError):
```
keithrfung, line 19: Want lots of feedback on these errors.

src/electionguard/discrete_log.py
```diff
@@ -6,16 +6,39 @@
...
 from .singleton import Singleton
-from .group import ElementModP, ONE_MOD_P, mult_p
+from .group import BaseElement, ElementModP, ONE_MOD_P, mult_p
 
-DLogCache = Dict[ElementModP, int]
-DLOG_MAX = 100_000_000
-"""The max number to calculate.  This value is used to stop a race condition."""
+DiscreteLogCache = Dict[ElementModP, int]
+
+_DLOG_MAX_EXPONENT = 100_000_000
```
keithrfung, line 13: This now becomes a backup to the backup. The first and foremost one can be set by using the ElectionGuard config. I expected lines like:

```py
DiscreteLog().set_max_exponent(config.max_votes)
DiscreteLog().precompute_cache()
```

src/electionguard/discrete_log.py
```diff
@@ -6,16 +6,39 @@
...
+
+
+class DiscreteLogExponentError(ValueError):
+    """Raised when the max exponent is larger than the system allows."""
+
+    def __init__(self, exponent: int, max_exponent: int = _DLOG_MAX_EXPONENT) -> None:
+        message = (
+            f"Discrete log exponent of {exponent} exceeds maximum of {max_exponent}."
+        )
+        super().__init__(message)
```
SteveMaier-IRT, line 26: The next error has the message directly in the init call.  We should be consistent and make the two match.

src/electionguard/discrete_log.py
```diff
@@ -57,30 +84,67 @@ async def discrete_log_async(
...
     exponent = cache[max_element]
 
     g = ElementModP(get_generator(), False)
+
     while element != max_element:
         exponent = exponent + 1
-        if exponent > DLOG_MAX:
-            raise ValueError("size is larger than max.")
+        if exponent > max_exponent:
+            raise DiscreteLogExponentError(exponent, max_exponent)
```
SteveMaier-IRT, line 145: Might be good to move this check right after line 138 to check the values before making the ElementModP.

src/electionguard/discrete_log.py
```diff
@@ -89,15 +153,41 @@ class DiscreteLog(Singleton):
...
+    _cache: DiscreteLogCache = {ONE_MOD_P: 0}
     _mutex = asyncio.Lock()
+    _max_exponent: int = _DLOG_MAX_EXPONENT
+    _lazy_evaluation: bool = True
+
+    def get_cache(self) -> DiscreteLogCache:
+        return self._cache
+
+    def set_max_exponent(self, max_exponent: int) -> None:
+        self._max_exponent = max_exponent
```
SteveMaier-IRT, line 165: If the new max that is set is < the current exponent then we should change the exponent like we do below.

src/electionguard/discrete_log.py
```diff
@@ -57,30 +84,67 @@ async def discrete_log_async(
...
     if not cache:
-        cache = {ONE_MOD_P: 0}
+        cache = _INITIAL_CACHE
+
     max_element = list(cache)[-1]
     exponent = cache[max_element]
 
     g = ElementModP(get_generator(), False)
+
     while element != max_element:
```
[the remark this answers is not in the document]
keithrfung: That actually will not work. This was the part that Lee and and I discussed. It is modular math. You cannot solve it this way. This is actually the reason for this PR.

src/electionguard/discrete_log.py
keithrfung: It adds a little bit of redundancy because we still need it here. Let's discuss.

SteveMaier-IRT (approved):
Changes and comments look good.

PR #562, opened 2022-04-04
💻 CLI V1

### LINKED ISSUE #561: ✨ Initial ElectionGuard CLI

### Suggestion

The initial version of the CLI should perform the exact same steps as tests/integration/test_end_to_end_election.py but from a command line.  It should have minimal human interaction.  Subsequent issues will flesh it out to add parameters and make it more usable.

### Possible Implementation

_No response_

### Anything else?

_No response_

lprichar:

### Issue
Fixes #561 

### Description
Adds a initial CLI program into election guard.  Thanks to [click](https://click.palletsprojects.com/en/8.1.x/) it contains command line documentation and the potential to have multiple commands.  

The code in this PR has two commands 1. a simple hello world to demonstrate how a second command will look.  2. an `e2e` command.

The e2e command will allow the running of an election including key ceremony, election building, submitting, encrypting, spoiling and casting votes, and tally and spoiled ballot decryption.  It does not yet publish an election record.  It currently takes two arguments: number of guardians and quorum.  It will eventually take a manifest and ballots, but not yet for this PR.

### Testing

1. Install the CLI per the readme.md document
2. Try `eg --help`

Expected: A list of the commands available

3. `eg e2e --help`

Expected: A description of the e2e command's parameters and usage

4. `eg e2e`

Expected: It should prompt you for missing parameters like quorum and number of guardians

5. `eg e2e --guardian-count=2 --quorum=2`

It should run through an election and print out results


keithrfung:
@lprichar Tried to help out. I have resolved the linting issue with #564 . Take a look at that. 
`click` had a conflict with the `black` linter. Updating that should resolve your issue.

I highly recommend you run `make auto-lint` after that. It will resolve any other `black` linting issues.

src/electionguard_cli/readme.md
```diff
@@ -0,0 +1,16 @@
+# Setup
```
keithrfung, line 1: This should actually be in the docs folder. That is deployed to py_pi

src/electionguard_cli/setup.py
```diff
@@ -0,0 +1,15 @@
+from setuptools import setup
```
keithrfung, line 1: This shouldn't be necessary. This is the other tool outside poetry. 
We should be wrapping this tool into our existing python package. 
Reach out to me about this.

src/electionguard_cli/readme.md
keithrfung, line 1: This should actual be in docs folder.

They are deployed to here:
https://pypi.org/project/electionguard
https://microsoft.github.io/electionguard-python/
https://electionguard-python.readthedocs.io/en/latest/

src/electionguard_cli/e2e_steps/submit_votes_step.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class SubmitVotesStep(E2eStepBase):
+    """Responsible for encrypting votes and storing them in a ballot store"""
+
+    def __encrypt_ballots(
+        self, plaintext_ballots: List[PlaintextBallot], encrypter: EncryptionMediator
+    ) -> List[CiphertextBallot]:
+        ciphertext_ballots: List[CiphertextBallot] = []
+        # Encrypt the Ballots
```
keithrfung, line 27: This comment seem redundant to method. The recommendation would be to include this within method name.

src/electionguard_cli/e2e_steps/submit_votes_step.py
```diff
@@ -0,0 +1,85 @@
...
+        self,
+        ballot_store: DataStore,
+        internal_manifest: InternalManifest,
+        context: CiphertextElectionContext,
+        ciphertext_ballots: List[CiphertextBallot],
+    ) -> None:
+        # Configure the Ballot Box
+        ballot_box = BallotBox(internal_manifest, context, ballot_store)
+
+        # todo: find a way for users to choose which ballot(s) to spoil
```
keithrfung, line 59: We try not to do TODO's in code and instead use issues with lines after the PR.

src/electionguard_cli/e2e_steps/submit_votes_step.py
```diff
@@ -0,0 +1,85 @@
...
+    def encrypt_votes(
+        self,
+        plaintext_ballots: List[PlaintextBallot],
+        internal_manifest: InternalManifest,
+        context: CiphertextElectionContext,
+    ) -> List[CiphertextBallot]:
+        self.print_header("Encrypting votes")
+        self.print_value("Loaded ballots", len(plaintext_ballots))
+
+        # Configure the Encryption Device
```
keithrfung, line 43: I'd make this a method.

src/electionguard_cli/e2e_steps/submit_votes_step.py
```diff
@@ -0,0 +1,85 @@
...
+        return self.__encrypt_ballots(plaintext_ballots, encrypter)
+
+    def cast_and_spoil(
+        self,
+        ballot_store: DataStore,
+        internal_manifest: InternalManifest,
+        context: CiphertextElectionContext,
+        ciphertext_ballots: List[CiphertextBallot],
+    ) -> None:
+        # Configure the Ballot Box
```
keithrfung, line 56: I'd make this a method.

src/electionguard_cli/e2e_steps/key_ceremony_step.py
```diff
@@ -0,0 +1,95 @@
...
+)
+from electionguard.guardian import Guardian
+from electionguard.key_ceremony_mediator import KeyCeremonyMediator
+from electionguard.utils import get_optional
+
+
+class KeyCeremonyStep(E2eStepBase):
+    """Responsible for running a key ceremony and producing an elgamal public key given a list of guardians."""
+
+    def __announce_guardians(
```
keithrfung, line 17: Is the double underscore `__` a click convention? This is usually used in certain methods like `__init__`, but I haven't seen it used like this?

https://docs.python.org/3/tutorial/classes.html#tut-private

src/electionguard_cli/e2e_steps/input_retrieval_step.py
```diff
@@ -0,0 +1,66 @@
...
+        self.print_value("Name", manifest_name)
+        self.print_value("Scope", manifest.election_scope_id)
+        self.print_value("Geopolitical Units", len(manifest.geopolitical_units))
+        self.print_value("Parties", len(manifest.parties))
+        self.print_value("Candidates", len(manifest.candidates))
+        self.print_value("Contests", len(manifest.contests))
+        self.print_value("Ballot Styles", len(manifest.ballot_styles))
+
+    def get_manifest(self) -> Manifest:
+        # todo: get manifest from params
```
keithrfung, line 36: Use issues instead of TODO's.

src/electionguard_cli/e2e_steps/input_retrieval_step.py
```diff
@@ -0,0 +1,66 @@
...
+from electionguard_tools.factories.election_factory import (
+    ElectionFactory,
+)
+
+
+class InputRetrievalStep(E2eStepBase):
+    """Responsible for retrieving and displaying user provided inputs"""
+
+    def get_ballots(self) -> List[PlaintextBallot]:
+        # todo: parameterize the plaintext ballot file
```
keithrfung, line 18: Use Issues instead of TODO's

src/electionguard_cli/e2e_steps/input_retrieval_step.py
```diff
@@ -0,0 +1,66 @@
...
+from electionguard.guardian import Guardian
+from electionguard.manifest import InternationalizedText, Manifest
+from electionguard_tools.factories.ballot_factory import BallotFactory
+from electionguard_tools.factories.election_factory import (
+    ElectionFactory,
+)
+
+
+class InputRetrievalStep(E2eStepBase):
+    """Responsible for retrieving and displaying user provided inputs"""
```
keithrfung, line 15: This is a nitpick, but more for informing. I believe the standard docstyle uses `.` at the end of these lines. I turned off the line formatting for this because it was annoying with some of existing files, but figured I'd catch if for the future.

src/electionguard_cli/e2e_steps/key_ceremony_step.py
keithrfung: This could also be dramatically simplified by using `electionguard_tools` and probably should be.

src/electionguard_cli/start.py
```diff
@@ -0,0 +1,9 @@
+import click
```
keithrfung, line 1: Does click require a `__main__` for startup?

keithrfung (changes requested):
`make auto-lint` has to be run. I checked while I was fixing pieces. 

I think some of the code could be reduced by using the existing code we have in `electionguard_tools`. There are some points that I couldn't mention on here that we may want to fix before pushing pr in. 

Particularly, this should be worked into build and package. This means we need to add this into pyproject.toml and it should be linted fully. Reach out to me and I can walk you through this. I think this is a pretty stellar start though.

src/electionguard_cli/e2e_steps/key_ceremony_step.py
lprichar: Correct, the "dunder" is used to indicate a private method since afaik Python doesn't have language support for private methods: https://www.geeksforgeeks.org/private-methods-in-python/

src/electionguard_cli/setup.py
lprichar: As discussed publishing the CLI so that end users can consume it via PIP Install is a separate issue that I created here: https://github.com/microsoft/electionguard-python/issues/568

src/electionguard_cli/readme.md
lprichar: Deleted the readme, as discussed

src/electionguard_cli/e2e_steps/key_ceremony_step.py
keithrfung: This isn't accurate. See this: https://www.youtube.com/watch?v=HTLu2DFOdTg&t=1988s

This doesn't make this actually private, but is actually localizing the function. The single `_` underscore might be closer to what you mean.

keithrfung: Here is the matching post for the same source you're using:
https://www.geeksforgeeks.org/underscore-_-python/

src/electionguard_cli/e2e_steps/key_ceremony_step.py
```diff
@@ -0,0 +1,95 @@
...
+            mediator.announce(guardian.share_key())
+        return get_optional(mediator.share_announced())
+
+    def __get_mediator(self, ceremony_details: CeremonyDetails) -> KeyCeremonyMediator:
+        mediator: KeyCeremonyMediator = KeyCeremonyMediator(
+            "mediator_1", ceremony_details
+        )
+        return mediator
+
+    def __share_Keys(
```
keithrfung, line 30: Random capital here.

keithrfung:
This should actually close #568 now btw

src/electionguard_cli/e2e_steps/key_ceremony_step.py
lprichar: ✅ dramatically simplified

PR #573, opened 2022-04-05
✅ Resolve Pylint not correctly linting all folders

### LINKED ISSUE #572: 🐞  Pylint not correctly linting

### Current Behavior

There seems to be an issue with how pylint does linting within Make. 

### Expected Behavior

_No response_

### Steps To Reproduce

`make lint` and `poetry run pylint ./src/**/*.py ./tests/**/*.py` produce different results
Screen Shot 2022-04-05 at 3 20 02 PM
Screen Shot 2022-04-05 at 3 20 51 PM

### Environment

```markdown
- OS:
```

### Anything else?

This card is to both fix this issue and to resolve the linting issues.

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #572 

### Description
- Pylint was not recognizing the folder paths. The direct path folder results in proper linting.
- Fixed floating unfixed errors.

### Testing
`make auto-lint` with no errors.


src/electionguard_tools/strategies/election.py
```diff
@@ -176,16 +176,6 @@ def reporting_unit_types(draw: _DrawType):
     return ReportingUnitType(n)
 
 
-@composite
-def annotated_emails(draw: _DrawType):
```
lprichar, line 180: deleting unused method?

lprichar (approved):
lg2m

src/electionguard_tools/strategies/election.py
keithrfung: It's worse than that. It's a duplicated method. Exactly the same named method.

src/electionguard_cli/e2e_steps/e2e_step_base.py
```diff
@@ -8,10 +8,13 @@ class E2eStepBase:
     from the CLI.
     """
 
+    header_color = "green"
+    value_color = "yellow"
+
     def print_header(self, s: str) -> None:
```
keithrfung, line 14: These functions to keep them in this file I added the header and value color as settable in the class. Truthfully, this should not be a base class like this. This should be a module with print functions. I will likely make an issue for this.

src/electionguard_cli/e2e_steps/input_retrieval_step.py
```diff
@@ -15,10 +15,6 @@
 class InputRetrievalStep(E2eStepBase):
     """Responsible for retrieving and displaying user provided inputs."""
 
-    def get_ballots(self) -> List[PlaintextBallot]:
```
keithrfung, line 18: The issue with these is all the same. They are not using `self` or using anything internal to the class. With the way python is structured everything is public, it doesn't make sense for this embedding and it makes complicated name structures. Almost all linters bark at this.

src/electionguard_tools/helpers/__init__.py
```diff
@@ -61,6 +61,7 @@
     "accumulate_plaintext_ballots",
     "export",
```
SteveMaier-IRT, line 62: Since export was renamed to export_record, do we still need the export name here?

keithrfung: This was the name of the module. These are automade

Issue #548, opened 2022-02-21
♻️  Refactor Words to line up better with ElectionGuard 1.0 Specification


keithrfung:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Suggestion

There are likely more words but one concern is that there are a couple key phrases that are confusing with the current specification. The goal is to refactor as minimally as possibly yet get closer to the specification wording. 

*Election Public Key** is a problematic phrase. The current python considers each of the keys the guardians combines as an election public key. These are unnamed in the specifications, however the election public key is used to describe the combined key used for encryption the election. This leads to difficulties surrounding the key shares (as referred to in specification) and the election partial key backups (as referred to in python). This is reduced substantially because of the removal of auxiliary keys and the need for guardians to have two different keys. 

### Possible Implementation

The main area of focus would be the Key Ceremony. 

The main refactors as denoted so far are:
`ElectionPublicKey` -> `GuardianKey` or `guardian_key` for parameters
`ElectionKeyPair` -> `GuardianKeyPair` or `guardian_key_pair` for parameters
`ElectionJointKey` -> `ElectionKey` or `election_key` for parameters
`ElectionPartialKeyVerification` -> `KeyShareVerification` or `verification` for parameters
`ElectionPartialKeyChallenge` -> `KeyShareChallenge` or `challenge` for parameters

Methods would be updated accordingly. 
The major piece is replacing `election_key_pair` -> `guardian_key_pair`

For example:
https://github.com/microsoft/electionguard-python/blob/c14758bc67ebbcdeb4f98958fbe4e0b0310a20f1/src/electionguard/key_ceremony.py#L268-L288

Refactored to
```py
def combine_guardian_keys(
    guardian_keys: List[GuardianKey],
) -> ElectionKey:
    public_keys = [set.key for set in guardian_keys]
    commitments = [
        commitment
        for set in guardian_keys
        for commitment in set.coefficient_commitments
    ]

    return ElectionKey(
        joint_public_key=elgamal_combine_public_keys(public_keys),
        commitment_hash=get_optional(
            hash_elems(commitments)
        ),  # H(K 1,0 , K 2,0 ... , K n,0 )
    )
```

### Anything else?

There may be other pieces to revisit. 

The major files should be 
`key_ceremony.py` and `guardian.py`

benaloh:
[...]

PR #583, opened 2022-04-07
♻️  Refactor key references to use elgamal types

### LINKED ISSUE #582: ♻️  Refactor key references to ElGamal key types

### Suggestion

There is a type for `ElGamalSecretKey` and `ElGamalPublicKey`. These give a lot of clarity for when and which keys are being passed in as opposed to which Element type. In addition, this shifts the dependency to elgamal module as opposed to the group module.

https://github.com/microsoft/electionguard-python/blob/ec69dec9cc802824ba37d273b3ffebdf7f6af9fc/src/electionguard/elgamal.py#L23-L24

### Possible Implementation

Implementations would be like this:

https://github.com/microsoft/electionguard-python/blob/ec69dec9cc802824ba37d273b3ffebdf7f6af9fc/src/electionguard/elgamal.py#L66

```py
def decrypt(self, secret_key: ElGamalSecretKey) -> int:
```

### Anything else?

Falls underneath #548 . This is being created to reduce scope on that issue.

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #582 

This relates to #548 

### Description
ElementModP is used instead of specifying ElGamalPublicKey type. This reduces the direct dependency on group and gives clarity with little changes.

### Testing
Existing test suite should be sufficient since this is just a rename.


lprichar (approved):
lg2m

PR #590, opened 2022-04-11
CLI Election Record Export

### LINKED ISSUE #567: ✨ CLI should publish election record

### Suggestion

The e2e command should take a parameter for where to output results.  If not specified it should not output results.  If it is specified then when it completes it should export a zip file with:

- manifest
- context
- constants
- lagrange coefficients
- plaintext ballots
- ciphertext ballots
- private guardian records

### Possible Implementation

_No response_

### Anything else?

_No response_

lprichar:
### Issue

Fixes #567 

### Description
You can now optionally provide a `--output-file="./election_record.zip"` parameter to the CLI's e2e command.  When you do so it will output a zip file at that location with a tally, manifest, context, constants, lagrange coefficients, guardians, submitted ballots, and everything else required to verify an election.

### Testing
1. Run `run_simple_election.sh`

It should output a file called `election_record.zip`


src/electionguard_cli/e2e_steps/election_record_step.py
```diff
@@ -0,0 +1,52 @@
...
+        decrypt_results: E2eDecryptResults,
+    ) -> None:
+
+        self.print_header("Election Record")
+        guardian_records = [
+            guardian.publish() for guardian in election_inputs.guardians
+        ]
+        constants = get_constants()
+        with tempfile.TemporaryDirectory() as temp_dir:
+            export_record(
```
lprichar, line 35: This file is the only real code in this PR, everything else is just moving stuff around in order to get access to all the bits of data that this step needs.

src/electionguard_cli/e2e_steps/election_record_step.py
```diff
@@ -0,0 +1,52 @@
...
+                [submit_results.device],
+                submit_results.data_store.all(),
+                decrypt_results.plaintext_spoiled_ballots.values(),
+                decrypt_results.ciphertext_tally.publish(),
+                decrypt_results.plaintext_tally,
+                guardian_records,
+                decrypt_results.lagrange_coefficients,
+                election_record_directory=temp_dir,
+            )
+            file_name = splitext(election_inputs.output_file)[0]
```
lprichar, line 48: `make_archive` requires a file without an extension, but it seemed kinda odd to ask the user to omit it, so I just strip whatever extension the user might have provided and let `make_archive` add it back.

src/electionguard_cli/e2e_steps/election_record_step.py
```diff
@@ -0,0 +1,52 @@
+from shutil import make_archive
+from os.path import splitext
+import tempfile
```
keithrfung, line 3: `from tempfile import TemporaryDirectory`

src/electionguard_cli/e2e_steps/election_record_step.py
```diff
@@ -0,0 +1,52 @@
...
+                decrypt_results.plaintext_spoiled_ballots.values(),
+                decrypt_results.ciphertext_tally.publish(),
+                decrypt_results.plaintext_tally,
+                guardian_records,
+                decrypt_results.lagrange_coefficients,
+                election_record_directory=temp_dir,
+            )
+            file_name = splitext(election_inputs.output_file)[0]
+            make_archive(file_name, self.COMPRESSION_FORMAT, temp_dir)
+            click.echo(
```
keithrfung, line 50: Similarly with the other comment, you can do:
`from click import echo`

It's a lot cleaner. This is not necessary change for this PR but comment for improvement overall.

src/electionguard_cli/e2e_steps/election_record_step.py
```diff
@@ -0,0 +1,52 @@
...
+from electionguard_cli.e2e_steps.e2e_step_base import E2eStepBase
+from electionguard.constants import get_constants
+
+from electionguard_tools.helpers.export import export_record
+
+
+class ElectionRecordStep(E2eStepBase):
+    """Responsible for publishing an election record after an election has completed."""
+
+    REMOVE_RAW_OUTPUT = True
```
keithrfung, line 17: These aren't being used?

src/electionguard_cli/e2e_steps/election_record_step.py
```diff
@@ -0,0 +1,52 @@
...
+    REMOVE_ZIP_OUTPUT = True
+    COMPRESSION_FORMAT = "zip"
```
keithrfung, line 19: If this doesn't have a `_` leading it can be changed. I think this may be able to be changed so this could just be a variable. 
`CompressionFormat`

src/electionguard_cli/e2e_steps/election_record_step.py
```diff
@@ -0,0 +1,52 @@
...
+    def run(
+        self,
+        election_inputs: E2eInputs,
+        build_election_results: BuildElectionResults,
+        submit_results: E2eSubmitResults,
+        decrypt_results: E2eDecryptResults,
+    ) -> None:
+
+        self.print_header("Election Record")
+        guardian_records = [
```
keithrfung, line 30: good use of comprehension

keithrfung (changes requested):

run_simple_election.sh
```diff
@@ -1 +1 @@
-eg e2e --guardian-count=2 --quorum=2 --manifest=data/election_manifest_simple.json --ballots=data/plaintext_ballots_simple.json --spoil-id=25a7111b-4334-425a-87c1-f7a49f42b3a2
\ No newline at end of file
+eg e2e --guardian-count=2 --quorum=2 --manifest=data/election_manifest_simple.json --ballots=data/plaintext_ballots_simple.json --spoil-id=25a7111b-4334-425a-87c1-f7a49f42b3a2 --output-file="./election_record.zip"
```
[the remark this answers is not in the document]
keithrfung:
```
simple-election:
       poetry run eg e2e --guardian-count=2 --quorum=2 --manifest=data/election_manifest_simple.json --ballots=data/plaintext_ballots_simple.json --spoil-id=25a7111b-4334-425a-87c1-f7a49f42b3a2 --output-file="./election_record.zip"
```

Then `make simple-election` 
Keeps with the current pattern and just as short.

src/electionguard_cli/e2e_steps/election_record_step.py
lprichar: Thanks, trying to be pydantic, but for the record I still think a map/reduce style like `guardian_records = election_inputs.guardians.map(g => g.publish())` is clearer.

run_simple_election.sh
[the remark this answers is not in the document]
lprichar: Yes, excellent point, so much better!

keithrfung:
The Mac Failure looks like a GitHub not a code issue. i can restart workflow if need be.

keithrfung (approved):

PR #597, opened 2022-04-15
Fix bug

### LINKED ISSUE #595: 🐞 Deserialized ElGamalCiphertext Objects + Scheduler = PicklingError

### Current Behavior

If you deserialize an object that has a property of type `ElGamalCiphertext` (e.g. SubmittedBallot) and then try to send it through a method that uses `Scheduler.schedule` (e.g. `CiphertextTally.batch_append`), then you get the following error: `_pickle.PicklingError: Can't pickle <class 'pydantic.dataclasses._Pydantic_ElGamalCiphertext_94140943706128'>: it's not the same object as pydantic.dataclasses._Pydantic_ElGamalCiphertext_94140943706128`

### Expected Behavior

You shouldn't get the error

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

lprichar:
### Issue

Fixes #595 

### Description
It used to be the case that if you deserialize an object that has a property of type ElGamalCiphertext (e.g. SubmittedBallot) and then try to send it through a method that uses Scheduler.schedule (e.g. CiphertextTally.batch_append), then you get the following error: _pickle.PicklingError: Can't pickle <class 'pydantic.dataclasses._Pydantic_ElGamalCiphertext_94140943706128'>: it's not the same object as pydantic.dataclasses._Pydantic_ElGamalCiphertext_94140943706128.  The error was because the elgamal.py class was using the wrong dataclass attribute.  This fixes it.

### Testing

Generally speaking something like this, although I had difficulty getting the unit test to pass because I kept getting an OverflowError.  If I can get this test to pass I'll include it with this PR.

```
from typing import List
from electionguard.scheduler import Scheduler
from electionguard.serialize import from_file
from electionguard.ballot import SubmittedBallot
from tests.base_test_case import BaseTestCase


class TestElGamal(BaseTestCase):
    """Tests related to the ElGamal object"""

    def accumulate(self, ballot) -> SubmittedBallot:
        return ballot

    def schedule(self, submitted_ballots: List[SubmittedBallot]):
        with Scheduler() as scheduler:
            ballots = [(b) for b in submitted_ballots]
            return scheduler.schedule(
                self.accumulate,
                ballots,
            )

    def test_elgamal_deserialize_and_schedule(self) -> None:
        ballot = from_file(SubmittedBallot, "tests/unit/data/submitted_ballot.json")
        self.schedule([ballot])
```

### DIFF
--- a/src/electionguard/elgamal.py
+++ b/src/electionguard/elgamal.py
@@ -1,5 +1,5 @@
-from pydantic.dataclasses import dataclass
 from typing import Any, Iterable, Optional, Union
+from pydantic.dataclasses import dataclass
 
 
 from .discrete_log import DiscreteLog
--- a/src/electionguard/elgamal.py
+++ b/src/electionguard/elgamal.py
@@ -1,5 +1,6 @@
+from dataclasses import dataclass as basedataclass
 from typing import Any, Iterable, Optional, Union
-from pydantic.dataclasses import dataclass
+from pydantic.dataclasses import dataclass as pydanticdataclass
 
 
 from .discrete_log import DiscreteLog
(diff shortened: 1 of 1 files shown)


src/electionguard/elgamal.py
```diff
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from pydantic.dataclasses import dataclass
```
keithrfung, line 1: Linting is failing for legit reason. The imports are in wrong order. 

`pydantic` is not a built-in module like `dataclasses` or `typing`

keithrfung (changes requested):

keithrfung:
This may not be this simple. This may be related to something a bit more.

lprichar:
The underlying pickling issue was resolved in https://github.com/microsoft/electionguard-python/pull/599, so I'm abandoning this PR

Issue #524, opened 2022-01-21
🐞 BigInts in constants.json are numbers but are strings in context.json


ramsdell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Big integers are represented as JSON numbers in constants.json, but as JSON strings that must be parsed into numbers in context.json.

### Expected Behavior

Please be consistent and pick one of the behaviors.

### Steps To Reproduce

Look at the JSON files.

### Environment

```markdown
- OS:
```

keithrfung:
This is currently a circular reference issue. The `BaseElement` which is used for all larger ints uses the `ElectionConstants` file. 

A simple solution is to export a different file like `ElectionConstantsRecord` to resolve this. I think in **ElectionGuard 2.0** we should rethink this dependency tree entirely.

ramsdell:
I looked carefully at all of the JSON files that make up an election record.  With the exception of constants.json, all of the files represent bignums as strings, not integers.  It would be better to export the constants using the same representation used in all the other files.  That way, the same JSON parser can be used for all of the JSON files.

keithrfung:
We ARE using the same JSON parser for Python, we are not using the same type. We're using pydantic which identifies those items as integers not bigints. In python, there is no difference. We have a class that is isolated to handle BigInts. 

Again, the issue is a circular reference. The big integer classes are currently all defined by being mod a large or a small prime. These constant files allow users to import the large or small prime that would define those big integer classes.

JohnLCaron:
On closer look, while I do see the quoted ElementMod vs unquoted BigInteger, my parser doesnt care. I was failing on the fact that the representation is now in base 10, before it was base 16.

Still need to nail this down in a formal spec.  I suppose base64 would be another possibility, for compactness.

keithrfung:
Closing this. See issue created to resolve this situation.

PR #599, opened 2022-04-18
🐍. Big Integer and Base Element

### LINKED ISSUE #598: 🐞  Align C++ and Python Large Integers

### Current Behavior

The never ending task. Resolve differences in C++ and Python serialization around large integers.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #598 

### Description
Back into the never ending serialization, the current push was to try and resolve the previous concerns around `{ data: ''}` in the BaseElement while also prepping for #524 . This takes a similar approach of using a value type to handle `BigInteger`. It also attempts to hide this a bit away from the group.py and isolate these concerns out. 

### Testing
The existing test suite is fairly thorough for these items.

Currently, this is ironically failing in the deserialization round of the end to end test. This may be a false positive because both dictionaries show identical values. The output of these can be seen easily by changing the output flags on the end to end test.

### DIFF
--- a/src/electionguard/big_integer.py
+++ b/src/electionguard/big_integer.py
@@ -12,10 +12,13 @@ def _hex_to_int(input: str) -> int:
 
 def _int_to_hex(input: int) -> str:
     """Given an int, returns a hex string representing bytes."""
+
+    def pad_hex(hex: str) -> str:
+        """Pad hex to ensure 2 digit hexadecimal format maintained."""
+        return "0" + hex if len(hex) % 2 else hex
+
     hex = format(input, "02X")
-    if len(hex) % 2:
-        hex = "0" + hex
-    return hex
+    return pad_hex(hex)
 
 
 _zero = mpz(0)
--- a/src/electionguard/big_integer.py
+++ b/src/electionguard/big_integer.py
@@ -1,0 +1,110 @@
+from typing import Any, Tuple, Union
+from base64 import b16decode
+
+# pylint: disable=no-name-in-module
+from gmpy2 import mpz
+
+
+def _hex_to_int(input: str) -> int:
+    """Given a hex string representing bytes, returns an int."""
+    return int(input, 16)
+
+
+def _int_to_hex(input: int) -> str:
+    """Given an int, returns a hex string representing bytes."""
+    hex = format(input, "02X")
+    if len(hex) % 2:
+        hex = "0" + hex
+    return hex
+
+
+_zero = mpz(0)
+
+
+def _convert_to_element(data: Union[int, str]) -> Tuple[str, int]:
+    """Convert element to consistent types"""
+    if isinstance(data, str):
+        hex = data
+        integer = _hex_to_int(data)
+    else:
+        hex = _int_to_hex(data)
+        integer = data
+    return (hex, integer)
+
+
+class BigInteger(str):
+    """A specialized representation of a big integer in python"""
+
+    _value: mpz = _zero
+
+    def __new__(cls, data: Union[int, str]):  # type: ignore
+        (hex, integer) = _convert_to_element(data)
+        big_int = super(BigInteger, cls).__new__(cls, hex)
+        big_int._value = mpz(integer)
+        return big_int
+
+    @property
+    def value(self) -> mpz:
+        """Get internal value for math calculations"""
+        return self._value
+
+    def __int__(self) -> int:
+        """Overload int conversion."""
+        return int(self.value)
+
+    def __eq__(self, other: Any) -> bool:
+        """Overload == (equal to) operator."""
+        return (
+            isinstance(other, BigInteger) and int(self.value) == int(other.value)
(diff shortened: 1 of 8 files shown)


src/electionguard/big_integer.py
```diff
@@ -0,0 +1,110 @@
...
+        ) or (isinstance(other, int) and int(self.value) > other)
+
+    def __ge__(self, other: Any) -> bool:
+        """Overload >= (greater than or equal) operator."""
+        return self.__gt__(other) or self.__eq__(other)
+
+    def __add__(self, other: Any) -> Any:
+        """Overload addition operator."""
+        if isinstance(other, str):
+            return str(self) + other
```
SteveMaier-IRT, line 88: this looks like it will concatenate the strings and add the values.

src/electionguard/big_integer.py
```diff
@@ -0,0 +1,110 @@
...
+        return self._value
+
+    def __int__(self) -> int:
+        """Overload int conversion."""
+        return int(self.value)
+
+    def __eq__(self, other: Any) -> bool:
+        """Overload == (equal to) operator."""
+        return (
+            isinstance(other, BigInteger) and int(self.value) == int(other.value)
```
SteveMaier-IRT, line 58: Do you want to add a check for a str type as well or will the BigInteger take care of that?

src/electionguard/big_integer.py
```diff
@@ -0,0 +1,110 @@
...
+
+def _hex_to_int(input: str) -> int:
+    """Given a hex string representing bytes, returns an int."""
+    return int(input, 16)
+
+
+def _int_to_hex(input: int) -> str:
+    """Given an int, returns a hex string representing bytes."""
+    hex = format(input, "02X")
+    if len(hex) % 2:
```
lprichar, line 16: Why pad numbers to an odd length?  I see this was happening previously, but seems like this might warrant a comment of explanation for why this is happening, and maybe method whose name documents what's happening like `hex = pad_to_even_length(hex)`, and/or maybe a dedicated unit test or two.

src/electionguard/big_integer.py
keithrfung: Yes. That's exactly what it does.

src/electionguard/big_integer.py
keithrfung: We could add one, but I don't know if we really need one. It would essentially be a hex checker.

tests/integration/test_end_to_end_election.py
```diff
@@ -534,7 +534,7 @@ def step_5_publish(self) -> None:
         make_archive(ELECTION_RECORD_DIR, ZIP_SUFFIX, ELECTION_RECORD_DIR)
         make_archive(PRIVATE_DATA_DIR, ZIP_SUFFIX, PRIVATE_DATA_DIR)
 
-        self.deserialize_data()
+        # self.deserialize_data()
```
lprichar, line 537: remove?

lprichar (approved):
didn't test it, but lg2m

tests/integration/test_end_to_end_election.py
keithrfung: This is the bug I was highlighting. I was demonstrating where it was.

src/electionguard/big_integer.py
keithrfung: I will make a note here.

src/electionguard/serialize.py
```diff
@@ -1,31 +1,39 @@
...
+from .group import ElementModP, ElementModQ
+from .proof import ProofUsage
 
 _T = TypeVar("_T")
 
 _indent = 2
 _encoding = "utf-8"
 _file_extension = "json"
 
+_config = Config(
```
keithrfung, line 23: This tool takes a config that will cast all the elements correctly. This solves for the particular problem where BigIntegers are brought in by strings on pydantic due to the lack of overrides. I don't really like pulling in another tool but at least this is top down and not bottom up.

src/electionguard/big_integer.py
SteveMaier-IRT: Wanted to make sure that was what was expected since the values will then NOT be the same as the string.

src/electionguard/big_integer.py
SteveMaier-IRT: It is actually padding if it IS an odd length to make sure that it has an even length in the end.

src/electionguard/serialize.py
SteveMaier-IRT: Agreed and if it gives the desired format that is worth it.

keithrfung:
The last two commits resolve the concerns. The arithmetic functions were removed for risk of confusion. This way only the `_p` or _q` math functions should be used or using the internal value.

SteveMaier-IRT (approved):
looks good.

PR #600, opened 2022-04-19
♻️ Refactor ElectionGuard Constants to BigInteger

### LINKED ISSUE #555: 📄 Change ElectionConstants big integers to store as hex

### Suggestion

Referencing issue #524, this is a quick solution to resolve that issue with minimal changes while we should revisit this as a whole in ElectionGuard 2.0.

### Possible Implementation

The `ElectionConstants` dataclass does not `ElementModP` and `ElementModQ` but instead int so these items serialize into integers. This can't be quickly solved by changing the types because the `ElementModP` and `ElementModQ` currently use this class to handle the modulus limits. 

Here is the Election Constants class
https://github.com/microsoft/electionguard-python/blob/c59fddcd89f85619ba801d6c38b1ea7f08210951/src/electionguard/constants.py#L8-L22

This is the fastest solution to resolve this problem that can also help highlight the differences in python as other languages. Create a `BigInteger` class that `BaseElement` inherits from. This class could be in a separate file called `.biginteger` that `ElementModQ` could inherit from then all the integers used in ElectionConstants could use this. 

```py
p: BigInteger
q: BigInteger
r: BigInteger
g: BigInteger
```

### Anything else?

_No response_

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #555 

### Description
This utilizes the new BigInteger class to handle exporting the ElectionConstants as a hex as well. 

### Testing
This can be seen by running the end to end test and looking at the constants output.

### DIFF
--- a/src/electionguard/constants.py
+++ b/src/electionguard/constants.py
@@ -4,21 +4,23 @@ from os import getenv
 from dataclasses import dataclass
 from enum import Enum
 
+from .big_integer import BigInteger
+
 
 @dataclass
 class ElectionConstants:
     """The constants for mathematical functions during the election."""
 
-    large_prime: int
+    large_prime: BigInteger
     """large prime or p"""
 
-    small_prime: int
+    small_prime: BigInteger
     """small prime or q"""
 
-    cofactor: int  # (p - 1) / q
+    cofactor: BigInteger  # (p - 1) / q
     """cofactor or r"""
 
-    generator: int
+    generator: BigInteger
     """generator or g"""  # 2^r mod p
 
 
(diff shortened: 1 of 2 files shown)


src/electionguard/constants.py
```diff
@@ -75,7 +77,7 @@ def get_constants() -> ElectionConstants:
     return option_map.get(option) or STANDARD_CONSTANTS
 
 
-get_large_prime = lambda: get_constants().large_prime
-get_small_prime = lambda: get_constants().small_prime
-get_cofactor = lambda: get_constants().cofactor
-get_generator = lambda: get_constants().generator
+get_large_prime = lambda: int(get_constants().large_prime.value)
```
keithrfung, line 80: I'm mildly concerned about the performance here. Any alternate version may be adding mpz to comparisons in BigInteger.

keithrfung: This could be a bug created for later if we wish and we could remove this.

lprichar (approved):
lg2m

Issue #592, opened 2022-04-14
⚡️ Add Election Setup Command to CLI


lprichar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Suggestion

Create a command `eg setup` to setup the election.

This command should use inputs to grab the details needed to complete the setup for an election so it is ready to encrypt ballots. This means it will have to complete a Key Ceremony (see `KeyCeremonyOrchestrator`) and create an Election Ciphertext Context to define the encryption context of an election. The outputs of this will be enough so that an external source can do encryption but also enough so the CLI can perform decryptions. 

An encryptor, a third party that does **encryption** will need:
📁  Encryption Package
- `Manifest.json`
- `CiphertextElectionContext.json` 
- `ElectionConstants.json`

For **decryption**, the information for the guardians that is present during the key ceremony is needed. 
The CLI is running an automated Key Ceremony so this will have to put out details like the `ElectionKeyPair` (Guardian's Key Pair) for the guardians. 

### Possible Implementation

Input: 
`Manifest.json`
Number of Guardians as text input
Quorum as text input

Output:
`CiphertextElectionContext.json` 
`ElectionConstants.json`
`GuardianRecord.json`
`GuardianKeyPair.json` <-- This is private info for decryption

Issue #574, opened 2022-04-06
✨ CLI Documentation


lprichar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Suggestion

Document the process of using the CLI

PR #606, opened 2022-04-21
🗳️ #586 CLI Import Ballots

### LINKED ISSUE #355: 🚧  Reconstruction of Guardian
## Original Issue

**Description**
The stateful `Guardian` class after the latest refactor is used to both run the Key Ceremony and create their decryption shares. Since these two pieces may exist over vastly different times this brings together the question of reconstructing the `Guardian`. The best way to proceed forward with this seems to allow a constructor or a factory method to reconstruct the guardians. 

~~This would need specifically to include most of the information in the `PrivateGuardianRecord`.~~

~~So two options to move forward on this:~~
~~1. Create a constructor with optionals to allow a guardian to be constructed with ElectionKeyPair, AuxiliaryKeyPair, sent and received ElectionPartialKeyBackups~~
~~2. Use a factory method to build a guardian and utilize the save and receive to consume a PrivateGuardianRecord to rebuild that guardian with the necessary information~~

------------------------------------------------------

## Updated Information 

When reconstructing the guardian it can be tempting to use the nonce, instead the correct methodology should be having the `ElectionKeyPair` as a parameter the `Guardian` instead of the nonce.

The nonce can be seen here:
https://github.com/microsoft/electionguard-python/blob/03698c683ab35f608e35dc9b66217caa1d40f7eb/src/electionguard/guardian.py#L157

But here we can notice that it is only used to create an `ElectionKeyPair`
https://github.com/microsoft/electionguard-python/blob/03698c683ab35f608e35dc9b66217caa1d40f7eb/src/electionguard/guardian.py#L177

### Implementation

There is a PrivateGuardianRecord that holds this info, but the simplest way is:
- [ ] Add a get_key_pair() to the Guardian and saving that file. 
- [ ] Remove the nonce seed from the init and replace with mandatory ElectionGuardKeyPair
- [ ] Use @classmethod to implement the nonce and other constructors.

The `get_key_pair` could eventually be changed to a `export_secret_key` if necessary to get the math value for saving on items like Yubikeys.

```py
class Guardian

    def __init__(
        self,
        id: str,
        sequence_order: int,
        number_of_guardians: int,
        quorum: int,
        key_pair: ElectionKeyPair,
    ) -> None:
        ...

    @classmethod
    def from_nonce(..., nonce: ElementModQ = None)
         """An optional `ElementModQ` value for nonce that can be used to generate the `ElectionKeyPair`. It is recommended to only use this field for testing"""
        generate_election_key_pair(id, sequence_order, quorum, nonce)
         return cls(..., key_pair)

    def get_key_pair(self) -> ElectionKeyPair:
          pass
 ```

### LINKED ISSUE #586: ✨ CLI should support importing submitted ballots

### Suggestion

Submitted ballots will be in a directory of ballot files which each contain cast/spoiled state in the file itself.

When used in this way the CLI should take as input:

- The directory with ballot files in it
- a manifest.json file
- A CiphertextElectionContext.json file
- An ElectionConstants.json file

### Possible Implementation

_No response_

### Anything else?

_No response_

### PR DESCRIPTION
**Pull request by lprichar**
### Issue

Fixes #586
Fixed #355 

### Description
Adds a new command to the CLI for importing and tallying ballots and decrypting spoiled ballots, and decrypting ballots.  This does not yet include election record, that functionality is in ticket https://github.com/microsoft/electionguard-python/issues/605.

### Testing

1. Run an end-to-end election like:

`poetry run eg e2e --guardian-count=2 --quorum=2 --manifest=data/election_manifest_simple.json --ballots=data/plaintext_ballots_simple.json --spoil-id=25a7111b-4334-425a-87c1-f7a49f42b3a2 --output-record="./election_record.zip" --output-keys="./guardian_private_data.json"`

2. Pull out the ballots, context, and manifest and plunk them in `\data\simple`
3. Run the import command like: `poetry run eg import-ballots --manifest=data/simple/manifest.json --context=data/simple/context.json --ballots-dir=data/simple/ballots --guardian-keys=./guardian_private_data.json` (or `make eg-import-simple-election`)

Expected results:

```
----------------------------------------
Retrieving Inputs
----------------------------------------

----------------------------------------
Retrieving manifest
----------------------------------------
Name: Jefferson County Spring Primary
Scope: jefferson-county-primary
Geopolitical Units: 4
Parties: 3
Candidates: 6
Contests: 2
Ballot Styles: 4
importing submitted_ballot_03a29d15-667c-4ac8-afd7-549f19b8e4eb.json
importing submitted_ballot_1048ce32-f1b1-4b05-b7fb-8c615ac842ee.json
importing submitted_ballot_25a7111b-4334-425a-87c1-f7a49f42b3a2.json
importing submitted_ballot_5a150c74-a2cb-47f6-b575-165ba8a4ce53.json
importing submitted_ballot_69aeacb4-64c6-4205-9bb2-5fb6b3b3ea58.json
importing submitted_ballot_9fee0e77-cfd2-401a-a210-93bbc4dd30ef.json
Ballots Dir: [...]/data/simple/ballots

----------------------------------------
Building election
----------------------------------------

----------------------------------------
Creating Tally
----------------------------------------
Ballots in tally: 6

----------------------------------------
Decrypting tally
----------------------------------------
Cast ballots: 5
Spoiled ballots: 1
Total ballots: 6
Guardian Present: 1
Guardian Present: 2
Lagrange coefficients retrieved: 2

----------------------------------------
Decrypted tally
----------------------------------------
Contest: justice-supreme-court
  john-adams-selection: 5 votes
  benjamin-franklin-selection: 0 votes
  john-hancock-selection: 2 votes
  write-in-selection: 3 votes
Contest: referendum-pineapple
  referendum-pineapple-affirmative-selection: 1 votes
  referendum-pineapple-negative-selection: 1 votes

----------------------------------------
Spoiled ballot '25a7111b-4334-425a-87c1-f7a49f42b3a2'
----------------------------------------
Contest: justice-supreme-court
  john-adams-selection: 1 votes
  benjamin-franklin-selection: 1 votes
  john-hancock-selection: 0 votes
  write-in-selection: 0 votes
  justice-supreme-court-4-placeholder: 0 votes
  justice-supreme-court-5-placeholder: 0 votes
```

### CODE REVIEW

File: data/simple/context.json [hunk 1]
```diff
@@ -0,0 +1,14 @@
+{
```
lprichar, line 1: I'm conflicted on whether to include this data in the `/data/simple` folder.  It made the `eg-import-simple-election` target work, right up until I added the `--guardian-keys=./guardian_private_data.json`, but then I thought it might be a bad idea to include a bunch of guardian private keys in the source code, even if it is just for a fake election.  So now I'm thinking I should just delete it all.  Thoughts? cc @keithrfung

src/electionguard/guardian.py
```diff
@@ -169,12 +176,25 @@ def __init__(
         self.id = id
         self.sequence_order = sequence_order
         self.set_ceremony_details(number_of_guardians, quorum)
-        self._backups_to_share = {}
-        self._guardian_election_public_keys = {}
-        self._guardian_election_partial_key_backups = {}
-        self._guardian_election_partial_key_verifications = {}
+        self._backups_to_share = {} if backups_to_share is None else backups_to_share
```
lprichar, line 179: I tried setting the default value of these new params to `{}` because it massively reduced cyclomatic complexity, but it produced linting errors, thus all these ternaries. I'd love to hear if there's a simpler way to do this.

src/electionguard/guardian.py
```diff
@@ -169,12 +176,25 @@ def __init__(
...
+            else election_partial_key_verifications
+        )
 
-        self.generate_election_key_pair(nonce_seed if nonce_seed is not None else None)
+        if election_keys is None:
+            self.generate_election_key_pair(
+                nonce_seed if nonce_seed is not None else None
+            )
+        else:
+            self._election_keys = election_keys
```
lprichar, line 197: All of the code in this file addresses #355.  I was going to make that a separate PR, but it was substantially simpler to know it was working correctly by including it in this CLI command, hope that's ok.

src/electionguard_cli/cli_models/e2e/e2e_inputs.py
```diff
@@ -2,10 +2,14 @@
...
 from electionguard.guardian import Guardian
 from electionguard.manifest import Manifest
+from electionguard_cli.cli_models.shared.cli_election_inputs_base import (
+    CliElectionInputsBase,
+)
 
 
-class E2eInputs:
-    """Responsible for holding the inputs to an end-to-end election."""
+# pylint: disable=too-many-instance-attributes
```
lprichar, line 10: This seems to be a very common lint rule to disable, maybe we should do it globally?

src/electionguard_cli/steps/shared/election_record_step.py
```diff
@@ -0,0 +1,82 @@
...
+        private_guardian_records = [
+            guardian.export_private_data() for guardian in election_inputs.guardians
+        ]
+        file_name = splitext(election_inputs.output_keys)[0]
+        file_path = dirname(election_inputs.output_keys)
+        to_file(private_guardian_records, file_name, file_path)
+        echo(f"Exported private guardian keys to '{election_inputs.output_keys}'")
+        file_basename = basename(election_inputs.output_keys)
+        self.print_warning(
+            f"{file_basename} can decrypt an entire election. Protect it. Encrypt it. Do not share it."
```
lprichar, line 81: I wanted a big red warning if you output all the private guardian keys.  Is this a reasonable approach?  cc @keithrfung

src/electionguard_cli/commands/e2e_command.py
```diff
@@ -1,12 +1,17 @@
 from io import TextIOWrapper
 import click
-from ..e2e_steps import (
-    ElectionBuilderStep,
+
+from electionguard_cli.steps.shared.tally_step import TallyStep
```
lprichar, line 4: will combine this import with the others

src/electionguard_cli/commands/e2e_command.py
```diff
@@ -45,43 +50,57 @@
...
+    build_election_results = ElectionBuilderStep().build_election_with_key(
         election_inputs, joint_key
     )
     submit_results = SubmitVotesStep().submit_votes(
         election_inputs, build_election_results
     )
-    decrypt_results = DecryptStep().decrypt_tally(
-        submit_results.data_store, election_inputs.guardians, build_election_results
+    (ciphertext_tally, spoiled_ballots) = TallyStep().get_from_ballot_store(
+        build_election_results, submit_results.data_store
```
lprichar, line 91: broke out tally work into it's own step since there was a lot of commonality between the two commands

src/electionguard_cli/steps/shared/election_record_step.py
```diff
@@ -0,0 +1,82 @@
...
+from electionguard_cli.cli_models import BuildElectionResults, E2eSubmitResults
+from electionguard_cli.cli_models.e2e_decrypt_results import E2eDecryptResults
+from electionguard_cli.cli_models.e2e.e2e_inputs import E2eInputs
+
+from electionguard_tools.helpers.export import export_record
+
+from .cli_step_base import CliStepBase
+
+
+class ElectionRecordStep(CliStepBase):
```
lprichar, line 17: Not much of this class actually changed, I just moved a lot of steps around and github seems to have failed to recognize the moves

src/electionguard_cli/steps/shared/print_results_step.py
```diff
@@ -0,0 +1,42 @@
...
+from electionguard.type import BallotId
+from electionguard.tally import (
+    PlaintextTally,
+)
+
+from electionguard_cli.cli_models import E2eDecryptResults
+from .cli_step_base import CliStepBase
+
+
+class PrintResultsStep(CliStepBase):
```
lprichar, line 12: Nothing in here changed

src/electionguard_cli/steps/shared/tally_step.py
```diff
@@ -0,0 +1,55 @@
...
+class TallyStep(CliStepBase):
+    """Responsible for creating a tally and retrieving spoiled ballots."""
+
+    def get_from_ballots(
+        self,
+        build_election_results: BuildElectionResults,
+        ballots: List[SubmittedBallot],
+    ) -> Tuple[CiphertextTally, List[SubmittedBallot]]:
+        self.print_header("Creating Tally")
+        tuble_ballots = [(None, b) for b in ballots]
```
lprichar, line 21: This code that converts a list of ballots into a list of tuples of ballots is awkward, but I was aiming to change the underlying python core as little as possible, and a list of tuples of ballots is what it currently takes, so I'm adapting this code to match.  I can change the source method if desired, I just worry about what other unintended consequences that may have.

src/electionguard_cli/steps/import_ballots/import_ballots_input_retrieval_step.py
```diff
@@ -0,0 +1,69 @@
...
+            guardian_keys, context
+        )
+        submitted_ballots = ImportBallotsInputRetrievalStep._get_ballots(ballots_dir)
+
+        self.print_value("Ballots Dir", ballots_dir)
+
+        return ImportBallotInputs(guardians, manifest, submitted_ballots, context)
+
+    @staticmethod
+    def _get_guardians(
```
lprichar, line 44: Here's where the magic happens.  This reconstructs each guardian including private keys from the guardian private key data that was exported when you append `--output-record="./election_record.zip"` to the e2e command.  Without something like this there would be no way to use the CLI to decrypt ballots that were encrypted in another system such as the C++ core.

src/electionguard_cli/steps/shared/election_record_step.py
```diff
@@ -0,0 +1,82 @@
...
+                decrypt_results.plaintext_tally,
+                guardian_records,
+                decrypt_results.lagrange_coefficients,
+                election_record_directory=temp_dir,
+            )
+            file_name = splitext(election_inputs.output_record)[0]
+            make_archive(file_name, self._COMPRESSION_FORMAT, temp_dir)
+            echo(f"Exported election record to '{election_inputs.output_record}'")
+
+    def _export_private_keys(self, election_inputs: E2eInputs) -> None:
```
lprichar, line 66: This method is new.  If you add `--output-record="./election_record.zip"` to the e2e command this method will export the private keys using the (existing) `guardian.export_private_data()` method so that when you run the `import-ballots` command you can reconstruct the private key data.  Technically the e2e command isn't exactly the right place to do this, it should go into the election setup command that will be built in #592, but it needs to be here for now because there's no other place.

data/simple/ballots/submitted_ballot_03a29d15-667c-4ac8-afd7-549f19b8e4eb.json
```diff
@@ -0,0 +1,185 @@
+{
```
keithrfung, line 1: These files are auto generated from the `electionguard` repository. These files should not be placed here.

Makefile
```diff
@@ -253,5 +253,9 @@ release-notes:
 	echo "## Issues" >> release_notes.md
 	curl "${GITHUB_API_URL}/${GITHUB_REPOSITORY}/issues?milestone=${MILESTONE_NUM}&state=all" | jq '.[].title' | while read i; do echo "[$i]($MILESTONE_URL)" >> release_notes.md; done
 
-simple-election:
-	poetry run eg e2e --guardian-count=2 --quorum=2 --manifest=data/election_manifest_simple.json --ballots=data/plaintext_ballots_simple.json --spoil-id=25a7111b-4334-425a-87c1-f7a49f42b3a2 --output-file="./election_record.zip"
+eg-e2e-simple-election:
+# to save keys to import append --output-keys="./guardian_private_data.json"
```
keithrfung, line 257: Is this an old todo?

data/simple/ballots/submitted_ballot_03a29d15-667c-4ac8-afd7-549f19b8e4eb.json
```diff
@@ -0,0 +1,185 @@
...
+  "object_id": "03a29d15-667c-4ac8-afd7-549f19b8e4eb",
```
keithrfung, line 2: These files are auto generated from the `electionguard` repository. These files should not be placed here.

Makefile
keithrfung, line 257: Is this an todo?

src/electionguard_cli/cli_models/e2e/e2e_inputs.py
keithrfung: We shouldn't. This is really strong code smell. We should be making adjustments to avoid it. For example, this should be a `dataclass`.

src/electionguard_cli/cli_models/import_ballots/import_ballot_inputs.py
```diff
@@ -0,0 +1,29 @@
...
+from electionguard.ballot import SubmittedBallot
+from electionguard.election import CiphertextElectionContext
+from electionguard.guardian import Guardian
+from electionguard.manifest import Manifest
+from electionguard_cli.cli_models.shared.cli_election_inputs_base import (
+    CliElectionInputsBase,
+)
+
+
+class ImportBallotInputs(CliElectionInputsBase):
```
keithrfung, line 11: For what it's worth, usually these models would be placed in the same module as the item using them. This seperation feels very C#.

```py
from random_step import RandomStepModel, random_step_method
```

src/electionguard_cli/cli_models/e2e/e2e_inputs.py
```diff
@@ -2,10 +2,14 @@
 from electionguard.ballot import PlaintextBallot
 from electionguard.guardian import Guardian
 from electionguard.manifest import Manifest
+from electionguard_cli.cli_models.shared.cli_election_inputs_base import (
```
keithrfung, line 5: incorrect import

src/electionguard_cli/start.py
```diff
@@ -1,11 +1,11 @@
 import click
-from electionguard_cli.commands import e2e, hello
+from electionguard_cli.commands import e2e, import_ballots
```
keithrfung, line 2: Could be relative import

src/electionguard_cli/steps/e2e/submit_votes_step.py
```diff
@@ -15,11 +15,15 @@
     ElectionFactory,
 )
 
-from ..cli_models import E2eInputs, BuildElectionResults, E2eSubmitResults
-from .e2e_step_base import E2eStepBase
+from electionguard_cli.cli_models.e2e.e2e_inputs import E2eInputs
```
keithrfung, line 18: use relative import

src/electionguard_cli/steps/import_ballots/import_ballots_input_retrieval_step.py
```diff
@@ -0,0 +1,69 @@
...
+from os.path import join
+from typing import List
+from click import echo
+from electionguard import CiphertextElectionContext
+
+from electionguard.ballot import SubmittedBallot
+from electionguard.guardian import Guardian, PrivateGuardianRecord
+from electionguard.manifest import Manifest
+from electionguard.serialize import from_file, from_file_wrapper, from_list_in_file
+from electionguard_cli.cli_models.import_ballots.import_ballot_inputs import (
```
keithrfung, line 12: use relative import

src/electionguard_cli/steps/shared/decrypt_step.py
```diff
@@ -3,19 +3,16 @@
 from electionguard_cli.cli_models import BuildElectionResults, E2eDecryptResults
```
keithrfung, line 3: Use relative import

src/electionguard_cli/steps/shared/decrypt_step.py
```diff
@@ -71,38 +71,21 @@ def decrypt_tally(
...
-def _get_decryption_mediator(
-    build_election_results: BuildElectionResults,
-) -> DecryptionMediator:
-    decryption_mediator = DecryptionMediator(
-        "decryption-mediator",
-        build_election_results.context,
-    )
-    return decryption_mediator
+    @staticmethod
+    def _compute_tally_share(
```
keithrfung, line 75: These methods feel overdone. They are just really wrapping existing methods with almost the same name.

src/electionguard_cli/steps/shared/election_builder_step.py
```diff
@@ -0,0 +1,53 @@
+from electionguard.elgamal import ElGamalPublicKey
+from electionguard.group import ElementModQ
+from electionguard.key_ceremony import ElectionJointKey
+from electionguard.election_builder import ElectionBuilder
+from electionguard.utils import get_optional
+from electionguard_cli.cli_models.e2e.e2e_inputs import CliElectionInputsBase
+
+from electionguard_cli.cli_models import BuildElectionResults
```
keithrfung, line 8: Relative imports

src/electionguard_cli/steps/shared/election_record_step.py
```diff
@@ -0,0 +1,82 @@
+from shutil import make_archive
+from os.path import splitext, dirname, basename
+from tempfile import TemporaryDirectory
+from click import echo
+from electionguard import to_file
+from electionguard.constants import get_constants
+
+from electionguard_cli.cli_models import BuildElectionResults, E2eSubmitResults
+from electionguard_cli.cli_models.e2e_decrypt_results import E2eDecryptResults
+from electionguard_cli.cli_models.e2e.e2e_inputs import E2eInputs
```
keithrfung, line 10: Relative imports

src/electionguard/guardian.py
```diff
@@ -154,6 +154,13 @@ def __init__(
         sequence_order: int,
         number_of_guardians: int,
         quorum: int,
+        election_keys: ElectionKeyPair = None,
+        backups_to_share: Dict[GuardianId, ElectionPartialKeyBackup] = None,
```
keithrfung, line 158: We should not have to preload the backups_to_share OR the election_partial_key_verifications. There is no use case for that.

src/electionguard/guardian.py
```diff
@@ -154,6 +154,13 @@ def __init__(
...
+        election_public_keys: Dict[GuardianId, ElectionPublicKey] = None,
+        partial_key_backups: Dict[GuardianId, ElectionPartialKeyBackup] = None,
+        election_partial_key_verifications: Dict[
+            GuardianId, ElectionPartialKeyVerification
+        ] = None,
         nonce_seed: Optional[ElementModQ] = None,
```
keithrfung, line 164: The nonce and the election_keys are in direct conflict with each other.

src/electionguard/guardian.py
keithrfung: This seems like it should have been isolated in its own PR.

keithrfung (changes requested):
This seems like it needs a lot of cleanup work. I think this is a workable branch for demo but I'd suggest a considerable amount of changes. 

- The sample data should be being retrieved from the `electionguard` repo.
- The guardian work should be isolated. 
- There's large code cleanup items. Lots and lots of @staticmethod where it's not necessary. inits where dataclass can be used and pretty wording coding.

Makefile
lprichar: Not a todo.  The objective was to add a helpful comment to anyone that might use it, since `output-keys` is a required input field for the `import-ballots` command, and you can only get one of those by outputting it in the e2e command like this.  If you definitely want to delete the sample data and the `eg-e2e-simple-election` command on line 256 then I'll remove this comment too, and add it to the description of https://github.com/microsoft/electionguard-python/issues/574.

src/electionguard_cli/cli_models/e2e/e2e_inputs.py
lprichar: We are currently disabling too-many-instance-attributes it in 12 other places in the code base.  I actually tried using the dataclass, but it wasn't smart enough to handle inheritance, I kept getting `TypeError: __init__() takes 5 positional arguments but 9 were given`.  I did a bit of research and it looks like Python 3.10 solves this with the `kw_only=true` dataclass attribute parameter: https://medium.com/@aniscampos/python-dataclass-inheritance-finally-686eaf60fbb5.  Any other thoughts on what I should be doing here?

src/electionguard_cli/cli_models/import_ballots/import_ballot_inputs.py
lprichar: The three top level folders in CLI are currently:

- `cli_models`
- `commands`, and
- `steps`

Just to confirm are you proposing eliminating the `cli_models` folder entirely and moving all of those objects into the `steps` folder?  That seems odd to me since the models are used by both commands and steps.  But if that's pythonic I'll certainly do it.

However, this case is especially difficult because ImportBallotInputs is used by classes in both `/steps/import_ballots`, and `steps/shared`.  Which of those two folders would you have me move this particular class into?

src/electionguard/guardian.py
lprichar: Fair enough, will try to do better in the future

src/electionguard/guardian.py
lprichar: It feels unbalanced to export data in `export_private_data` that isn't imported in `from_private_record`.  How would you feel if instead I were to remove `backups_to_share` `election_partial_key_verifications` from both?  It's currently only used by `TestEndToEndElection`, and I don't think it would break anything there to remove them there.

src/electionguard_cli/steps/e2e/submit_votes_step.py
lprichar: I'm happy to switch it to a module import, but I've been having two problems with relative imports:

1. Do you ever get these linting errors in Visual Studio when you attempt to use relative imports that go up a directory?

image

If so, what do you do about them?

2. When I `make mypy` with any `from ...` commands I get the following linting errors:

`/steps/e2e/submit_votes_step.py:18:1: error: Relative import climbs too many namespaces`

Should I look into disabling that mypy linting error?

src/electionguard_cli/steps/shared/decrypt_step.py
lprichar: That goes up three levels and produces a mypy linting error, see prior comment

src/electionguard_cli/steps/e2e/submit_votes_step.py
lprichar: I was able to solve issue #1 by deleting `/electionguard_cli/__init__.py`.  That also solved another linting problem I was having.  Still need guidance on issue #2.

src/electionguard_cli/steps/shared/election_record_step.py
lprichar: That's up three, will produce linting error, see other comment

src/electionguard_cli/steps/shared/election_builder_step.py
lprichar: That would go up three and producing linting error, see prior comment

src/electionguard/guardian.py
keithrfung: That's fine. That's a straightforward solution.

keithrfung (approved):

src/electionguard/guardian.py
lprichar: That is true.  Users would need to either not supply the final optional argument and/or use named parameters.  If that is undesirable, I would need guidance on how to fix, since as far as I know Python doesn't support constructor overloading, which is how I would address it in a different language.

src/electionguard_cli/steps/shared/election_record_step.py
```diff
@@ -43,8 +59,24 @@ def run(
...
+        ]
+        file_path = election_inputs.output_keys
+        for private_guardian_record in private_guardian_records:
+            file_name = private_guardian_record.guardian_id
+            to_file(private_guardian_record, file_name, file_path)
+        echo(
+            f"Exported {len(private_guardian_records)} private guardian keys to '{election_inputs.output_keys}'"
+        )
+        self.print_warning(
+            f"The files in {file_path} can decrypt an entire election. Protect the folder securely and do not share it."
```
keithrfung, line 81: The files in {file_path} are secret and should be protected securely and not shared publicly.

data/simple/context.json
keithrfung: The data should not be included here but PR'd and included at the electionguard folder. Anything in the election record is fine to have available. I'd actually suggest making a full election record and having that in the data. There is one that could be replaced.

keithrfung (changes requested):
Waiting on removal of sample data and rebase

keithrfung (approved):

Issue #607, opened 2022-04-22
🤔  Revisit export of private guardian information


keithrfung:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Suggestion

The `PrivateGuardianRecord` is designed to be part of the data schema for one particular use case.

https://github.com/microsoft/electionguard-python/blob/03698c683ab35f608e35dc9b66217caa1d40f7eb/src/electionguard/guardian.py#L88-L110

This class is designed to be an export of exclusively the private information of the Guardian. A class that could perhaps be saved on a yubikey or similar device. It should only have the information that the Guardian should maintain for the decryption process and nothing else. 

### Possible Implementation

Simplify the class down to less items and include the decrypted values of the backups as opposed to the encrypted. This will only matter once secret sharing is complete but it means the class is inaccurate at this juncture. 

```
@dataclass
class PrivateGuardianRecord
        guardian_id: GuardianId
        election_keys: ElectionKeyPair
        guardian_key_shares: Dict[GuardianId, ElementModQ] # Same as decrypted_backups, could be a class
```

### Anything else?

Closely relates to #355

PR #608, opened 2022-04-22
🚧 #355 Guardian Reconstruction

### LINKED ISSUE #355: 🚧  Reconstruction of Guardian

lprichar:
### Issue

Fixes #355 

### Description
Allows creating a guardian via `PrivateGuardianRecord`

### Testing
The unit test `test_import_from_guardian_private_record` in `tests/unit/test_guardian.py` should pass


src/electionguard/guardian.py
```diff
@@ -95,20 +95,12 @@ class PrivateGuardianRecord:
     election_keys: ElectionKeyPair
     """Private election Key pair of this guardian"""
 
-    backups_to_share: Dict[GuardianId, ElectionPartialKeyBackup]
```
keithrfung, line 98: See this issue. The private record that is produced should have just the raw values and the not the entire backup. 

This will also change how the constructor works and what information is saved. 

https://github.com/microsoft/electionguard-python/issues/607

keithrfung (changes requested):

src/electionguard/guardian.py
keithrfung: The private record builder looks solid but the nonce should still be isolated into its own class method `from_nonce`. The nonce + election key is a redundant constructor.

lprichar: Please clarify: I don't understand what action the first comment requires, or if that's just background information.  

As far as the second comment about Nonce I took a stab at it here: https://github.com/microsoft/electionguard-python/pull/608/commits/f8cb5d0b82a44b04b0501274460aaaf92b3c0edb.  I don't love that `generate_key_pair` parameter.  I can remove it if you don't mind that `generate_election_key_pair` would get called twice when the constructor is called from the `from_nonce` static method.  

Incidentally, I would normally solve this kind of thing by having two constructors, one with nonce, one with keys, but I couldn't figure out how to do that.  Not sure what I'm missing, but would welcome guidance if this approach isn't what you're looking for.

lprichar: Per our conversation I:
- Switched from staticmethods to classmethods for constructors
- Made a from_nonce overload
- Restored all removed fields from PrivateGuardianRecord
- Moved `generate_election_key_pair` out of the constructor

To accomplish that last point, I had to make a new `@classmethod` type constructor which I called `from_context_info` for lack of a better name that takes neither nonce, nor key data, but then calls `generate_election_key_pair`.  I then refactored all code in the codebase that creates a guardian using the old constructor and switched it to `Guardian.from_context_info(...)`.

src/electionguard/guardian.py
```diff
@@ -154,7 +154,13 @@ def __init__(
         sequence_order: int,
         number_of_guardians: int,
         quorum: int,
-        nonce_seed: Optional[ElementModQ] = None,
+        election_keys: ElectionKeyPair = None,
```
keithrfung, line 157: Refactor to `key_pair`. This should be required.

src/electionguard/guardian.py
```diff
@@ -154,7 +154,13 @@ def __init__(
         sequence_order: int,
```
keithrfung, line 154: `id` and`sequence_order` are no longer necessary. The ElectionKeyPair includes both these values.

src/electionguard/guardian.py
```diff
@@ -163,18 +169,82 @@ def __init__(
...
+            {}
+            if guardian_election_partial_key_verifications is None
+            else guardian_election_partial_key_verifications
+        )
+
+        if not election_keys is None:
+            self._election_keys = election_keys
+
+    @classmethod
+    def from_context_info(
```
keithrfung, line 196: This could be rolled into the `from_nonce`. No reason for an almost duplicate constructor.

src/electionguard/guardian.py
```diff
@@ -163,18 +169,82 @@ def __init__(
...
+        """Creates a guardian with an `ElementModQ` value that will be used to generate
+        the `ElectionKeyPair`. This method should generally only be used for testing."""
+
+        guardian = cls(
+            id,
+            sequence_order,
+            number_of_guardians,
+            quorum,
+        )
+        guardian.generate_election_key_pair(nonce_seed)
```
keithrfung, line 225: No reason to use this method forcing construction immediately. Directly, use the `generate_election_key_pair` function and in addition, remove the original function.

src/electionguard/guardian.py
```diff
@@ -163,18 +169,82 @@ def __init__(
...
+        private_guardian_record: PrivateGuardianRecord,
+        number_of_guardians: int,
+        quorum: int,
+    ) -> "Guardian":
+        guardian = cls(
+            private_guardian_record.guardian_id,
+            private_guardian_record.election_keys.sequence_order,
+            number_of_guardians,
+            quorum,
+            private_guardian_record.election_keys,
```
keithrfung, line 240: This constructor is only going to be used AFTER a key ceremony.  In which case there isn't need to save all this extra information. These are the only necessities to save since they are used in threshold decryption.

```
private_guardian_record.election_keys
private_guardian_record.guardian_election_public_keys,
private_guardian_record.guardian_election_partial_key_backups
```

src/electionguard/guardian.py
keithrfung, line 196: This could be rolled into the `from_nonce`. No reason for an almost duplicate constructor.

src/electionguard/guardian.py
keithrfung, line 157: Refactor to `key_pair`. This field should be required.

src/electionguard/guardian.py
keithrfung, line 154: `id` and`sequence_order` are no longer necessary. The ElectionKeyPair includes both these values.

src/electionguard/guardian.py
```diff
@@ -163,18 +169,82 @@ def __init__(
...
+    def from_nonce(
+        cls,
+        id: str,
+        sequence_order: int,
+        number_of_guardians: int,
+        quorum: int,
+        nonce_seed: Optional[ElementModQ] = None,
+    ) -> "Guardian":
+        """Creates a guardian with an `ElementModQ` value that will be used to generate
+        the `ElectionKeyPair`. This method should generally only be used for testing."""
```
keithrfung, line 217: Can add a note here. Random nonce will be generated if none provided. This is what generate election key pair does.

keithrfung (changes requested):
Close a little refactoring to simplify.

src/electionguard/guardian.py
keithrfung: To expand on this, there are 3 constructors already.

1.  `__init__` , which uses a provided `ElectionKeyPair`
2. `from_nonce`, which generates an ElectionKeyPair using a nonce (or random nonce if none provided)
3. `from_record`, which is post decryption

src/electionguard/guardian.py
lprichar: Except that ElectionKeyPair is optional.  The vast majority of instantiations of Guardian use `from_context_info` and thus do not provide a key pair.

src/electionguard/guardian.py
lprichar: Will refactor rename, but it cannot be made required without significant modifications because the vast majority of instantiations of Guardian do not provide a key pair.

src/electionguard/guardian.py
lprichar: The problem is that this constructor is the one used by the majority of instantiations of Guardian.  It can't be rolled into `from_nonce` because callees don't have a nonce, and it can't be rolled into `from_private_record` because callees don't have a `PrivateGuardianRecord`.  

It could go back into `__init__` if you like, but then we'd have to move the `generate_election_key_pair` call back into there, and put back the if statement so that the other constructors don't call it.  Should I do that?

src/electionguard/guardian.py
keithrfung: Incorrect. The ElectionKeyPair is never optional. A guardian cannot exist without one. The majority of the existing ones are using the constructor internally. This merely makes it more obvious by pushing it into a specific constructor.

keithrfung: You're overcomplicating it.

```
def __init__(
      self,
      key_pair: ElectionKeyPair,
      ceremony_details: CeremonyDetails
) -> None:
      self.id = key_pair.id
      self.sequence_order = key_pair.sequence_order
      self.ceremony_details = ceremony_details

# Applies to all existing cases as they currently are implemented.  
@classmethod
def from_nonce(
        cls, 
        id: str,
        sequence_order: int,
        number_of_guardians: int,
        quorum: int,
        nonce_seed: Optional[ElementModQ] = None):
        key_pair = generate_election_key_pair(
            self.id, self.sequence_order, self.ceremony_details.quorum, nonce
        )
        cls(key_pair, CeremonyDetails(number_of_guardians, quorum))
```

keithrfung: I just wrote up an example. It even is able to remove the lint skip. 
https://github.com/microsoft/electionguard-python/tree/feature/355-gaurdian-reconstruction-example

src/electionguard/guardian.py
keithrfung: In these cases, they ARE using a nonce. This is a misnomer hidden by the code. They will randomly generate a nonce when none is available. 

Either the key_pair is provided OR a nonce will be used to create one.

https://github.com/microsoft/electionguard-python/blob/03698c683ab35f608e35dc9b66217caa1d40f7eb/src/electionguard/election_polynomial.py#L78

lprichar: merged

docs/1_Key_Ceremony.md
```diff
@@ -70,7 +73,7 @@ guardians: List[Guardian]
 # Setup Guardians
 for i in range(NUMBER_OF_GUARDIANS):
   guardians.append(
-    Guardian(f"some_guardian_id_{str(i)}", i, NUMBER_OF_GUARDIANS, QUORUM)
+    Guardian.from_context_info(f"some_guardian_id_{str(i)}", i, NUMBER_OF_GUARDIANS, QUORUM)
```
keithrfung, line 76: .

keithrfung (approved):

PR #609, opened 2022-04-25
✨ Padding and Encoding

### LINKED ISSUE #612: ✨ Contest Data and Encoding

### Suggestion

Contest Data and Padding are a prerequisites for issues regarding #385 and secret sharing. This also should help center some of the encoding information to a single source. 

### Possible Implementation

_No response_

### Anything else?

_No response_

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #612 

### Description
This PR aims to add the underlying padding functionality necessary for secret sharing, overvotes and writeins. There are a couple sections to focus on so please dive into comments. 

**Core**
- Add a padding tool for encoding usage specifically with ElGamal Hashing. 
- Create an indicator for the data size of contests and selections
- Add settings for storing configuration.

**Boyscouting**
- Isolate encoding and byte order into a single file. 

### Testing
There is a new test `test_encode.py`

### DIFF
--- a/src/electionguard/election.py
+++ b/src/electionguard/election.py
@@ -5,6 +5,7 @@ from typing import Dict, Optional
 
 from .constants import get_small_prime, get_large_prime, get_generator
 from .elgamal import ElGamalPublicKey
+from .encode import PaddedDataSize
 from .group import (
     ElementModQ,
     ElementModP,
@@ -25,6 +26,18 @@ class Configuration:
     """
     Maximum votes, the maximum votes allowed on a selection for an aggregate ballot or tally.
     This can also be seen as the maximum ballots where a selection on a ballot can only have one vote.
+    """
+
+    write_in_size: PaddedDataSize = field(default=PaddedDataSize.Bytes_64)
+    """
+    The size of a write-in that can be stored in a selection.
+    This is a padded data size so the value represents the space available for storage.
+    """
+
+    contest_error_size: PaddedDataSize = field(default=PaddedDataSize.Bytes_32)
+    """
+    The size of a error that can be stored in the contest.
+    This is a padded data size so the value represents the space available for storage.
     """
 
 
--- a/src/electionguard/encode.py
+++ b/src/electionguard/encode.py
@@ -1,0 +1,50 @@
+"""Converting module that define padding and default settings for byte order and byte encoding."""
+
+from enum import IntEnum
+
+
+BYTE_ORDER = "big"
+BYTE_ENCODING = "utf-8"
+
+PAD_INDICATOR_SIZE = 2
+_PAD_BYTE = b"\x00"
+
+
+class PaddedDataSize(IntEnum):
+    """Define the sizes for padded data."""
+
+    Bytes_32 = 32 - PAD_INDICATOR_SIZE
+    Bytes_64 = 64 - PAD_INDICATOR_SIZE
+    Bytes_128 = 128 - PAD_INDICATOR_SIZE
+    Bytes_256 = 256 - PAD_INDICATOR_SIZE
+    Bytes_512 = 512 - PAD_INDICATOR_SIZE
+
+
+class TruncationError(ValueError):
+    """A specific truncation error to indicate when padded data is truncated."""
+
+
+def add_padding(
+    message: bytes, size: PaddedDataSize, allow_truncation: bool = True
+) -> bytes:
+    """Add padding to message in bytes."""
+    message_length = len(message)
+    if message_length > size:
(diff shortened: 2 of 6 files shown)


src/electionguard/election.py
```diff
@@ -27,6 +28,18 @@ class Configuration:
     This can also be seen as the maximum ballots where a selection on a ballot can only have one vote.
     """
 
+    write_in_size: PaddedDataSize = field(default=PaddedDataSize.Bytes_64)
```
keithrfung, line 31: These are names I came up with to describe these fields and what they do that I thought would be useful.

It could be strongly argued for these fields to be called.

`selection_extended_data_size`
`contest_extended_data_size`

src/electionguard/encode.py
```diff
@@ -0,0 +1,50 @@
...
+BYTE_ENCODING = "utf-8"
+
+PAD_INDICATOR_SIZE = 2
+_PAD_BYTE = b"\x00"
+
+
+class PaddedDataSize(IntEnum):
+    """Define the sizes for padded data."""
+
+    Bytes_32 = 32 - PAD_INDICATOR_SIZE
```
keithrfung, line 16: This was the most creative way I could come up with to showcase the relationship of the padding to the padding indicator size. This may be an argument where it more clever than understandable.

src/electionguard/encode.py
```diff
@@ -0,0 +1,50 @@
+"""Converting module that define padding and default settings for byte order and byte encoding."""
+
+from enum import IntEnum
+
+
+BYTE_ORDER = "big"
+BYTE_ENCODING = "utf-8"
```
keithrfung, line 7: These were scattered through some files. This consolidates them into one.

tests/unit/test_encoding.py
```diff
@@ -0,0 +1,55 @@
...
+            self.assertEqual(size + PAD_INDICATOR_SIZE, len(padded))
+            self.assertEqual(input, output)
+
+        except TruncationError:
+            padded = add_padding(byte_format, size, True)
+            unpadded = remove_padding(padded, size)
+            output = unpadded.decode(BYTE_ENCODING)
+
+            self.assertEqual(size + PAD_INDICATOR_SIZE, len(padded))
+            self.assertNotEqual(input, output)
```
keithrfung, line 55: This is intuitive even though it may not appear to be BECAUSE this indicates the text would have been truncated so the length is correctly padded but the input != output.

src/electionguard/encode.py
keithrfung: These values would show up in the json as 30, 62, etc. as one would expect.

src/electionguard/encode.py
```diff
@@ -0,0 +1,50 @@
...
+    Bytes_256 = 256 - PAD_INDICATOR_SIZE
+    Bytes_512 = 512 - PAD_INDICATOR_SIZE
+
+
+class TruncationError(ValueError):
+    """A specific truncation error to indicate when padded data is truncated."""
+
+
+def add_padding(
+    message: bytes, size: PaddedDataSize, allow_truncation: bool = True
```
keithrfung, line 28: I defaulted to allow truncation as the group had decided but this setting could be added to config quite quickly. Valid suggestion.

keithrfung: `allow_write_in_truncation` for example

src/electionguard/election.py
lprichar: The name you have seems solid, I consider it a good sign when names match descriptions, which this does

tests/unit/test_encoding.py
```diff
@@ -0,0 +1,55 @@
...
+    add_padding,
+    remove_padding,
+    PaddedDataSize,
+)
+
+
+class TestEncoding(TestCase):
+    """Guardian tests"""
+
+    def test_byte_padding(self) -> None:
```
lprichar, line 16: This is a very comprehensive unit test.  That said, I personally find it hard to read and hard to understand.  In my experience the most useful unit tests are small, have exactly one reason to fail, and have names with specificity.

In particular if I've broken someone else's unit test I want the name to tell me exactly what went wrong so I can fix it without having to read and understand a lot of context.

Of course the real world isn't always that easy, and this would probably be more work than worth to refactor into many smaller tests.  So let's leave it as a very comprehensive test for now, but maybe keep an eye out for opportunities to simplify in the future.

lprichar (approved):

keithrfung:
Found out this can't be merged as is. Going to revisit.

PR #613, opened 2022-04-27
🗳️ #592 election setup

### LINKED ISSUE #592: ⚡️ Add Election Setup Command to CLI

### LINKED ISSUE #602: 🐞  Remove unnecessary Key Ceremony Steps

### Current Behavior

In this section of code, the full ceremony is already performed before a second set of rounds is done. This isn't required. 
https://github.com/microsoft/electionguard-python/blob/03698c683ab35f608e35dc9b66217caa1d40f7eb/src/electionguard_cli/e2e_steps/key_ceremony_step.py#L24-L30

### Expected Behavior

Only the `perform_full_ceremony` should be necessary. The other steps can be removed. 

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

lprichar:
### Issue

Fixes #592 
Fixes #602

### Description
Adds a `setup` command to the CLI which will run an automated key ceremony and produces the files necessary to encrypt ballots, decrypt an election, and produce an election record.

### Testing
1. `poetry run eg setup --help`

Expected output:

```
Usage: eg setup [OPTIONS]

  This command runs an automated key ceremony and produces the files necessary
  to both encrypt ballots, decrypt an election, and produce an election
  record.

Options:
  --guardian-count INTEGER  The number of guardians that will participate in
                            the key ceremony and tally.
  --quorum INTEGER          The minimum number of guardians required to show
                            up to the tally.
  --manifest FILENAME       The location of an election manifest.
  --out DIRECTORY           The location of a directory into which will be
                            placed the output files such as context,
                            constants, and guardian keys. Existing files will
                            be overwritten.
  --help                    Show this message and exit.
```

2. `make eg-setup-simple-election`

Expected output:

```
poetry run eg setup --guardian-count=2 --quorum=2 --manifest=data/election_manifest_simple.json  --out=../data/out

----------------------------------------
Retrieving Inputs
----------------------------------------
Name: Jefferson County Spring Primary
Scope: jefferson-county-primary
Geopolitical Units: 4
Parties: 3
Candidates: 6
Contests: 2
Ballot Styles: 4

----------------------------------------
Performing key ceremony
----------------------------------------
Joint Key: A6..E4

----------------------------------------
Building election
----------------------------------------
Initializing public key and commitment hash
Creating context and internal manifest

----------------------------------------
Generating Output
----------------------------------------
Context: /mnt/c/dev/ElectionGuard/data/out.json
Election constants: /mnt/c/dev/ElectionGuard/data/out.json
Guradian records: /mnt/c/dev/ElectionGuard/data/out/guardians
Guardian private keys: /mnt/c/dev/ElectionGuard/data/out/guardian_private_data
WARNING: The files in /mnt/c/dev/ElectionGuard/data/out/guardian_private_data are secret and should be protected securely and not shared publicly.
```

And the files should all be present as promised


src/electionguard_cli/cli_steps/election_builder_step.py
```diff
@@ -28,13 +29,15 @@ def _build_election(
     ) -> BuildElectionResults:
         self.print_header("Building election")
 
+        click.echo("Initializing public key and commitment hash")
```
lprichar, line 32: For some reason the ElectionBuilder constructor takes a very long time to run on my machine, and the build command not so speedy either, so added some text so the user isn't left wondering.

src/electionguard_cli/cli_steps/input_retrieval_step_base.py
```diff
@@ -31,3 +32,12 @@ def get_first_value(text: Optional[InternationalizedText]) -> str:
         self.print_value("Candidates", len(manifest.candidates))
         self.print_value("Contests", len(manifest.contests))
         self.print_value("Ballot Styles", len(manifest.ballot_styles))
+
+    @staticmethod
+    def _get_guardians(number_of_guardians: int, quorum: int) -> List[Guardian]:
```
lprichar, line 37: This was just in e2e, but now setup uses it too, so I promoted it to the base class

src/electionguard_cli/cli_steps/key_ceremony_step.py
```diff
@@ -22,11 +22,6 @@ def run_key_ceremony(self, guardians: List[Guardian]) -> ElectionJointKey:
             "mediator_1", guardians[0].ceremony_details
         )
         KeyCeremonyOrchestrator.perform_full_ceremony(guardians, mediator)
```
lprichar, line 24: `perform_full_ceremony` does all the same stuff as the 3 commands below, so removed duplicate logic.  May need to switch back to the individual steps when we get the CLI to perform a key ceremony in steps, but can cross that bridge when we get to it.

src/electionguard_cli/cli_steps/input_retrieval_step_base.py
keithrfung: This should exist in a helper in the `electionguard_tools` if you want to really promote it. 
This line is already used in there to create the gaurdians in the KeyCeremonyOrchestrator. 

Would reduce repetition overall.

keithrfung: See this already existing static method.
https://github.com/microsoft/electionguard-python/blob/f0d6f01400d2e71feb7cc1e51af64aa202fa9097/src/electionguard_tools/helpers/key_ceremony_orchestrator.py#L11

src/electionguard_cli/setup_election/setup_election_command.py
```diff
@@ -0,0 +1,53 @@
...
+    This command runs an automated key ceremony and produces the files
+    necessary to encrypt ballots, decrypt an election, and produce an election record.
+    """
+
+    # get user inputs
+    election_inputs = SetupInputRetrievalStep().get_inputs(
+        guardian_count, quorum, manifest, out
+    )
+
+    # perform election
```
keithrfung, line 48: This comment seems off. The election is when voters are voting.

src/electionguard_cli/setup_election/setup_election_command.py
```diff
@@ -0,0 +1,53 @@
...
+)
+def SetupElectionCommand(
+    guardian_count: int, quorum: int, manifest: TextIOWrapper, out: str
+) -> None:
+    """
+    This command runs an automated key ceremony and produces the files
+    necessary to encrypt ballots, decrypt an election, and produce an election record.
+    """
+
+    # get user inputs
```
keithrfung, line 43: Comment redundant with code.

src/electionguard_cli/cli_steps/output_step_base.py
```diff
@@ -0,0 +1,47 @@
...
+        for private_guardian_record in private_guardian_records:
+            file_name = GUARDIAN_PREFIX + private_guardian_record.guardian_id
+            to_file(private_guardian_record, file_name, file_path)
+        self.print_value("Guardian private keys", output_keys)
+        self.print_warning(
+            f"The files in {file_path} are secret and should be protected securely and not shared publicly."
+        )
+
+    @staticmethod
+    def _get_guardian_records(
```
keithrfung, line 32: This is only used publicly but is named privately

keithrfung: Also this seems like a method to just put on the inputs base?

src/electionguard_cli/cli_steps/output_step_base.py
```diff
@@ -0,0 +1,47 @@
...
+        election_inputs: CliElectionInputsBase,
+    ) -> List[GuardianRecord]:
+        return [guardian.publish() for guardian in election_inputs.guardians]
+
+    def _export_file(
```
keithrfung, line 37: This should be a public method and definitely placed in its own module. Perhaps an export module that just does this. Maybe even in `electionguard_tools`

keithrfung (changes requested):
Small changes. See notes

src/electionguard_cli/cli_steps/output_step_base.py
lprichar: I really don't know what to do with this, and it's a pattern I'm struggling with constantly in Python.  In most OO languages it'd be a protected virtual instance method that the inheriting classes would call on their base class.  The inheriting classes might even choose to override some of that functionality if they wanted.  So that's why I don't consider this usage scenario to be public, because the two inheriting classes `ElectionRecordStep` and `OutputSetupFilesStep` are just calling it on their parent class `OutputStepBase`.

However, if I make `_get_guardian_records` an instance method then the linter complains `Method could be a function (no-self-use)`.  But if I solve that by making it static then the base classes can't override it, and worse it's suddenly considered public in Python?  And once it's static then it's considered a code smell.  

Any guidance would be greatly appreciated, keeping in mind this is just one instance of a pattern I'm running into frequently, so the consequences of this conversation reach much farther than this one instance.

src/electionguard_cli/cli_steps/output_step_base.py
lprichar: But it's specific to the CLI because of the `self.print_value` call.  Do you want me to just move lines 45 and 46 into `electionguard_tools`?

src/electionguard_cli/e2e/election_record_step.py
```diff
@@ -30,7 +29,7 @@ def export(
         self._export_election_record(
             election_inputs, build_election_results, submit_results, decrypt_results
         )
-        self._export_private_keys(election_inputs)
+        self._export_private_keys_e2e(election_inputs)
```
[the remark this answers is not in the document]
lprichar: The answer generally is because I'm an adherent of Clean Code by Bob Martin, in which he argues for small methods with as few parameters as possible.  The answer specifically is because the class `ElectionRecordStep` is a class that's specific to the E2E command (I renamed it to make that clearer in the next PR), and so it feels reasonable to accept a single parameter that is also specific to the E2E command as a convenience method that then calls the more generic parent class.  However, I don't feel particularly strongly about it in this instance and would be happy to delete the convenience method if you would prefer now that you understand my reasoning.

src/electionguard_cli/cli_steps/output_step_base.py
keithrfung: This doesn't reference anything in this class. I would put this on election_inputs since that is actually what is using this and simply add it as a getter. 

```py
class ElectionInputs()
        ...
        def get_election_records(self) -> List[GuardianRecord]:
                return [guardian.publish() for guardian in self.guardians]
```

keithrfung: Otherwise this code can be just left as is.

src/electionguard_cli/cli_steps/output_step_base.py
keithrfung: The file stuff and location has slowly been creeping in to all these setups. I'm not a huge fan. The to_file already does the location concatenation within that function. That could just return the location. Then that line wouldn't be necessary.

src/electionguard_cli/cli_steps/output_step_base.py
lprichar: True, although ElectionInputs is currently a dataclass.  Doesn't the existing architecture favor anemic domain models?  I don't mind going against that, I'm just trying to understand when to use that pattern and when not to, and I'm especially interested because this problem seems to keep cropping up for me.

keithrfung (approved):

PR #616, opened 2022-04-29
✨ Contest Data and Encoding

### LINKED ISSUE #612: ✨ Contest Data and Encoding

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #612 

### Description
- Add ContestData and ContestErrorType
- Migrate byte settings into utils
- Ensure that contest data can be consistently serialized and deserialized before encryption for overvotes and writeins. 

### Testing
There is a new test in `test_encrypt.py`. This should be verified.

### DIFF
--- a/src/electionguard/serialize.py
+++ b/src/electionguard/serialize.py
@@ -14,6 +14,7 @@ from .ballot_box import BallotBoxState
 from .manifest import ElectionType, ReportingUnitType, VoteVariationType
 from .group import ElementModP, ElementModQ
 from .proof import ProofUsage
+from .utils import BYTE_ENCODING, BYTE_ORDER, ContestErrorType
 
 _T = TypeVar("_T")
 
@@ -70,7 +71,7 @@ def from_file_wrapper(type_: Type[_T], file: TextIOWrapper) -> _T:
 def from_file(type_: Type[_T], path: Union[str, Path]) -> _T:
     """Deserialize json file as type."""
 
-    with open(path, "r", encoding=_encoding) as json_file:
+    with open(path, "r", encoding=BYTE_ENCODING) as json_file:
         data = json.load(json_file)
     return from_dict(type_, data, _config)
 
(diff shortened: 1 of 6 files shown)


src/electionguard/utils.py
```diff
@@ -1,11 +1,23 @@
...
 from base64 import b16decode
 
 _T = TypeVar("_T")
 _U = TypeVar("_U")
 
+BYTE_ORDER = "big"
+BYTE_ENCODING = "utf-8"
+
+
+class ContestErrorType(Enum):
```
keithrfung, line 14: Many of these were placed here due to circular dependencies due to the config in the serialize file.

src/electionguard/serialize.py
```diff
@@ -48,13 +99,15 @@ def construct_path(
...
 
 
-def from_raw(type_: Type[_T], obj: Any) -> _T:
-    """Deserialize raw as type."""
+def from_raw(type_: Type[_T], raw: Union[str, bytes]) -> _T:
+    """Deserialize raw json string as type."""
 
-    return parse_raw_as(type_, obj)
+    if type_ is datetime:
+        return parse_raw_as(type_, raw)
```
keithrfung, line 106: This is tech debt. There are several tests that specifically work with datetime to test the deserialization but it's not placed in as a json format. This should become and issue to fix up those tests.

src/electionguard/serialize.py
```diff
@@ -37,6 +44,50 @@
 )
 
 
+class PaddedDataSize(IntEnum):
+    """Define the sizes for padded data."""
+
+    Bytes_512 = 512 - PAD_INDICATOR_SIZE
```
keithrfung, line 50: There is only one size now but this will likely change in the future. I could be convinced to modify this.

PR #621, opened 2022-05-03
Adding electionguard tools to package

### LINKED ISSUE #578: 🐞  Include electionguard_tools in package

### Current Behavior

The electionguard_tools should be included within the package since electionguard_cli depends on them. 

### Expected Behavior

Add electionguard_tools to this section
https://github.com/microsoft/electionguard-python/blob/8aa589c9744bdf52e1e2659e2f83c255ffc65441/pyproject.toml#L23-L27

Ensure `pylint` and `mypy` are also checking electionguard_tools

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

MatthewRatzell:

### Issue

"kinda" Fixes #578 

### Description
I added the election_guard tools to be checked by mypy; however, of the 66 mypy errors that came up, I was only able to comfortably fix 44 of them.  The other issues require changing the return type of certain functions and I am not comfortable enough with the codebase to do this because the last thing I want to do is brick a function. Hopefully, the 44 errors I did fix were done correctly as I am new to Python and typing in python, in general, seems very different than what I am used to.

### Testing
Running the "make mypy" rule will show the remaining errors


src/electionguard_tools/factories/ballot_factory.py
```diff
@@ -2,6 +2,7 @@
 import os
 from random import Random, randint
 import uuid
+from xmlrpc.client import Boolean
```
keithrfung, line 5: This is the incorrect import for this section.

keithrfung (changes requested):
This is a really good start. We're gonna have to fix all the mypy changes to ensure this works and merge it though.

src/electionguard_tools/strategies/election.py
```diff
@@ -1,7 +1,8 @@
+from ast import Str
```
keithrfung, line 1: This import is also incorrect.

keithrfung:
@MatthewRatzell We're still waiting on followup on this PR.

src/electionguard_tools/factories/ballot_factory.py
MatthewRatzell: I went ahead and removed this improper import as well as the other improper input.

MatthewRatzell:
Super sorry about the late response, I'll have time tmmr to try and flush out some more of the mypy errors, and I removed those two improper imports if I need to replace them with something else I will. I also hope I resolved those two conversations correctly, I'm still getting used to the whole PR process

PR #622, opened 2022-05-04
✨  Write-In and Overvote Encryption

### LINKED ISSUE #385: ✨  Ensure Overvotes can be passed into system

### LINKED ISSUE #508: 🔐  Encrypt extended_data on contest and selection

### Suggestion

On a ballot, both the contest and selection should have an `extended_data` field:
https://github.com/microsoft/electionguard-python/blob/37b55504b62c07cd6083ddcbebb8c6af84e574e6/src/electionguard/ballot.py#L74

On the selection this allows for data to be stored like write-ins for the contest this can help cover cases like issues like #385. This information should be encrypted using hashed elgamal.  

References issue #47 

### Possible Implementation

This card should include both adding the encryption, the field and ensuring that the encryption and decryption function

- [ ] Add extended_data field to Contest for Ballot
- [ ] Add decryption of extended_data for spoiled ballots
- [ ] Add tests
- [ ] Add documentation where necessary
- [ ] Replace ElgamalCiphertext with HashedElgamalCiphertext for extended_data fields

### Anything else?

_No response_

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #385 
Fixes #508 

### Description
Encryption of write-ins and overvotes. This does **NOT** include the share decryption. 

The methodology of this changes how the original intent of encrypting write-ins and overvotes was. 
To reduce the fields encrypted and thus the amount of exponentiations, contest data is encrypted at the top level which includes the write-ins. 

The `extended-data` field on the contest is now a `HashedElGamalCiphertext` which is an encrypted byte of the `ContestData` which contains both the write-ins and any overvote information. For thoroughness, this PR also captures a null vote and an undervote even though these also pass through the system. 

### Testing
Existing test suite plus additional tests regarding `valid` should be considered. 
This is primarily focused on only the encryption.


lprichar (changes requested):
This is such an important part of the system I feel it should really have multiple tests, especially in validation, but it should at least have one that covers an overvote.

src/electionguard/ballot.py
```diff
@@ -313,10 +306,83 @@ class PlaintextBallotContest(OrderedObjectBase):
...
+        if self.total_selected > description.number_elected:
+            raise OverVoteException(self.object_id, self.selected_ids)
+
+        if description.votes_allowed is not None:
+            if self.total_votes > description.votes_allowed:
+                raise OverVoteException(self.object_id, self.selected_ids)
+
+            # Support for other cases such as cumulative voting not currently supported.
+            # (individual selections being an encryption of > 1)
+            if self.total_selected < description.votes_allowed:
```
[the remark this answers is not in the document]
keithrfung: Will handle this with `@cached_property
https://docs.python.org/3.8/library/functools.html?highlight=#functools.cached_property

keithrfung:
> This is such an important part of the system I feel it should really have multiple tests, especially in validation, but it should at least have one that covers an overvote.

I'm planning on adding tests anyhow. The existing test suite didn't even pass. These tests will be redundant with exist is_valid tests so will just work those into that.

src/electionguard/encrypt.py
```diff
@@ -404,21 +411,20 @@ def encrypt_contest(
...
-        return encrypted_contest
+    if should_verify_proofs or not encrypted_contest.proof:
+        if encrypted_contest.is_valid_encryption(
+            contest_description_hash, elgamal_public_key, crypto_extended_base_hash
+        ):
+            return encrypted_contest
+        log_warning(
+            f"mismatching contest proof for contest {encrypted_contest.object_id}"
+        )
+        return None
```
keithrfung, line 425: These changes are to try avoid the default case being `return None` and minimize the number of `return`s

tests/unit/test_encrypt.py
```diff
@@ -49,3 +92,122 @@ def _padding_cycle(self, data: ContestData) -> None:
...
+        contest_data = ContestData.from_bytes(decrypted_data)
+
+        # Assert
+        self.assertIsNotNone(contest_data)
+        self.assertIsNotNone(contest_data.error)
+        self.assertIsNotNone(contest_data.error_data)
+        self.assertEqual(contest_data.error, ContestErrorType.OverVote)
+        self.assertGreater(len(contest_data.error_data), 0)
+
+    def test_contest_encrypt_with_write_ins(self):
```
keithrfung, line 170: Write-In Test

tests/unit/test_encrypt.py
```diff
@@ -49,3 +92,122 @@ def _padding_cycle(self, data: ContestData) -> None:
...
+        # Assert
+        self.assertIsNotNone(encrypted_contest)
+        if encrypted_contest is not None:
+            self.assertTrue(
+                encrypted_contest.is_valid_encryption(
+                    contest_hash, keypair.public_key, encryption_seed
+                )
+            )
+
+    def test_contest_encrypt_with_overvotes(self) -> None:
```
keithrfung, line 124: Overvote Test

tests/unit/test_ballot.py
```diff
@@ -0,0 +1,92 @@
...
+        "favorite-star-wars-character",
+        ballot_selections,
+        None,
+        None,
+        placeholder_selections,
+    )
+    return description
+
+
+class TestBallot(BaseTestCase):
```
keithrfung, line 41: Tests as requested by PR review

src/electionguard/encrypt.py
```diff
@@ -256,7 +263,6 @@ def encrypt_selection(
     return None
 
 
-# pylint: disable=too-many-return-statements
```
keithrfung, line 259: This error could be removed by collapsing some of the other code.

src/electionguard/election.py
```diff
@@ -16,7 +16,7 @@
 class Configuration:
     """Configuration of election to allow edge cases."""
 
-    allow_overvotes: bool = field(default=False)
+    allow_overvotes: bool = field(default=True)
```
keithrfung, line 19: This setting is not currently passed through. This will eventually be hooked up. It will however require that information from the `Context` is passed through all the encrypt functions. Likely should be it's own issue following the PR. This is exactly how it is working on C++

src/electionguard/election_object_base.py
```diff
@@ -30,3 +30,15 @@ class OrderedObjectBase(ElectionObjectBase):
 def sequence_order_sort(unsorted: List[_Orderable_T]) -> List[_Orderable_T]:
     """Sort by sequence order."""
     return sorted(unsorted, key=lambda item: item.sequence_order)
+
+
+def list_eq(
```
keithrfung, line 35: Circular dependency issues with the placement of this function originally.

src/electionguard/elgamal.py
```diff
@@ -113,10 +113,10 @@ class HashedElGamalCiphertext:
     pad: ElementModP
     """pad or alpha"""
 
-    data: bytes
+    data: str
```
keithrfung, line 116: These could be `BigInteger` but the truth is they are just hexadecimal strings. `bytes.fromhex` is used instead of the more built in [`hex`](https://docs.python.org/3/library/functions.html#hex) due to some of the unique logic. 

Long term, the better plan would be to remove any of this customization and just use built in python methods.

src/electionguard/encrypt.py
```diff
@@ -321,7 +327,11 @@ def encrypt_contest(
         # false is entered instead and the selection_count is not incremented
         # this allows consumers to only pass in the relevant selections made by a voter
         for selection in contest.ballot_selections:
-            if selection.object_id == description.object_id:
+            # If overvote, no votes should be counted and instead placeholders should be used.
+            if (
+                selection.object_id == description.object_id
+                and error is not ContestErrorType.OverVote
```
keithrfung, line 333: This is a quick sidestep to avoid dramatically changing the loops in this section. 

Essentially, if an over vote has been detected, none of the selections from the ballot are appended. This means empty selections as per the descriptions from the manifest will be used instead.

src/electionguard/utils.py
```diff
@@ -14,11 +16,53 @@
 class ContestErrorType(Enum):
     """Various errors that can occur on ballots contest after voting."""
 
+    Default = "default"
     NullVote = "nullvote"
     UnderVote = "undervote"
     OverVote = "overvote"
 
 
+class ContestException(Exception):
```
keithrfung, line 25: In an ideal world, these would not exist in this file. However, the amount of circular dependencies that arose in this PR were troublesome. This file allowed some sidestep.

tests/unit/test_encrypt.py
```diff
@@ -49,3 +92,122 @@ def _padding_cycle(self, data: ContestData) -> None:
...
+        decrypted_data = get_optional(
+            encrypted_contest.extended_data.decrypt(keypair.secret_key, encryption_seed)
+        )
+        contest_data = ContestData.from_bytes(decrypted_data)
+
+        # Assert
+        self.assertIsNotNone(contest_data)
+        self.assertIsNotNone(contest_data.write_ins)
+        if contest_data is not None and contest_data.write_ins is not None:
+            self.assertGreater(len(contest_data.write_ins), 0)
```
lprichar, line 213: Why not `for selection in contest_data: self.assertEqual("write_in", selection.write_in)` to match the set up on line 182?

tests/unit/test_encrypt.py
lprichar: 💖

lprichar (approved):
lg2m!

tests/unit/test_encrypt.py
keithrfung: Can add.

PR #627, opened 2022-05-04
🐞  Resolve datetime deserialization issue

### LINKED ISSUE #628: 🐞  Datetimes are incorrectly deserialized

### Current Behavior

datetime fromisoformat is not designed to take strings such as `2020-03-01T13:00:00Z`

### Expected Behavior

Use an alternate dateparser such as `dateutils`

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: Any
```

### Anything else?

```
from dateutil import parser

example = parser.parse('2020-03-01T13:00:00Z')
```

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #628 

### Description
`datetime` was being deserialized with the incorrect isoformat function. Replaces this with a datetime call to ensure the datetimes are more consistent. 

### Testing
An existing manifest check was modified since this is the prime location where this is actually used.


tests/unit/test_manifest.py
```diff
@@ -55,16 +56,31 @@ def test_manifest_has_deterministic_hash(self) -> None:
     def test_manifest_hash_is_consistent_regardless_of_format(self) -> None:
 
         # Act
+        @dataclass
+        class DateType:
+            """Temp date class for testing"""
+
+            date: datetime
```
keithrfung, line 63: The tool only works on json which makes sense. This ensures there is an actual class and removes the tech debt around `from_raw`

lprichar (approved):
lg2m

PR #631, opened 2022-05-05
🔐 #617 CLI Encrypt Ballots

lprichar:
### Issue

Fixes #617 

### Description
Adds a new command to the CLI `encrypt-ballots` that will:

1. read a context, manifest, and a folder of plaintext ballots 
2. encrypt the ballots using the information from the context (and manifest)
3. export the encrypted ballots and a device file to a location specified

### Testing
Run something like this: `poetry run eg encrypt-ballots --ballots-dir=../data/encrypt/ballots --manifest=../data/encrypt/manifest.json --context=../data/encrypt/context.json --out-dir=../data/encrypt-out`

It should output something like this:

```
----------------------------------------
Retrieving Inputs
----------------------------------------
Name: Jefferson County Spring Primary
Scope: jefferson-county-primary
Geopolitical Units: 4
Parties: 3
Candidates: 6
Contests: 2
Ballot Styles: 4
Importing 03a29d15-667c-4ac8-afd7-549f19b8e4eb.json
Importing 1048ce32-f1b1-4b05-b7fb-8c615ac842ee.json
Importing 25a7111b-4334-425a-87c1-f7a49f42b3a2.json
Importing 5a150c74-a2cb-47f6-b575-165ba8a4ce53.json
Importing 69aeacb4-64c6-4205-9bb2-5fb6b3b3ea58.json
Importing 9fee0e77-cfd2-401a-a210-93bbc4dd30ef.json
ballot #1 object_id: 03a29d15-667c-4ac8-afd7-549f19b8e4eb

----------------------------------------
Building election
----------------------------------------
Initializing public key and commitment hash
Creating context and internal manifest

----------------------------------------
Encrypting Ballots
----------------------------------------
Ballots to encrypt: 6
Device location: polling-place
About to encrypt 6 ballots
Encrypting ballot: 03a29d15-667c-4ac8-afd7-549f19b8e4eb
Encrypting ballot: 1048ce32-f1b1-4b05-b7fb-8c615ac842ee
Encrypting ballot: 25a7111b-4334-425a-87c1-f7a49f42b3a2
Encrypting ballot: 5a150c74-a2cb-47f6-b575-165ba8a4ce53
Encrypting ballot: 69aeacb4-64c6-4205-9bb2-5fb6b3b3ea58
Encrypting ballot: 9fee0e77-cfd2-401a-a210-93bbc4dd30ef

----------------------------------------
Writing Encrypted Ballots
----------------------------------------
Device file: [dir]/data/encrypt-out/device.json.json
[...]

PR #633, opened 2022-05-06
💥 Prevent name conflicts with CLI commands

### LINKED ISSUE #617: ✨ CLI Encrypt Ballots

### Suggestion

Ability to encrypt ballots via the CLI.  This does not include submitting them, that is #585

### Possible Implementation

_No response_

### Anything else?

_No response_

lprichar:
### Issue

Fixes #617 

### Description
This addresses a comment in PR https://github.com/microsoft/electionguard-python/pull/631 about potential collisions of names in the Python repo.  Addressing it creates a lot of churn (29 files changed), so I broke it out into a separate PR.  However, all this does is prefix all folders in the CLI with `cli_`.

### Testing
Run the commands and ensure they work e.g. `make eg-e2e-simple-election`


keithrfung (changes requested):
A simpler approach is remove the recursive line in the make auto-lint and remove the lower level __init__ files. Don't expose the folders and commands. The only thing truly used from the CLI is really the setup.py

PR #635, opened 2022-05-10
🐛 Import guardians in all files in a folder not in a file

### LINKED ISSUE #634: 🐞 CLI Import Ballots Should Import Guardians within a Folder

### Current Behavior

The import ballots command imports guardians in a file, but it should import them in a directory to match how they're output by the election setup command.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

lprichar:
### Issue

Fixes #634 

### Description
The guardian import was looking in a file not for all files in a folder, this fixes that.

### Testing
Run something like `poetry run eg import-ballots --manifest=../data/import/manifest.json --context=../data/import/context.json --ballots-dir=../data/import/ballots --guardian-keys=../data/import/guardian_private_data --encryption-device=../data/import/device_91755434160.json --output-record=../data/import-publish-results.zip`

### DIFF
--- a/src/electionguard_cli/import_ballots/import_ballots_input_retrieval_step.py
+++ b/src/electionguard_cli/import_ballots/import_ballots_input_retrieval_step.py
@@ -1,14 +1,14 @@
 from typing import List
 from io import TextIOWrapper
 from os import listdir
-from os.path import isfile, isdir, join
+from os.path import join
 
 from electionguard import CiphertextElectionContext
 from electionguard.ballot import SubmittedBallot
 from electionguard.encrypt import EncryptionDevice
 from electionguard.guardian import Guardian, PrivateGuardianRecord
 from electionguard.manifest import Manifest
-from electionguard.serialize import from_file, from_list_in_file
+from electionguard.serialize import from_file
 
 from .import_ballot_inputs import (
     ImportBallotInputs,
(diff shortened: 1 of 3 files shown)


src/electionguard_cli/import_ballots/import_ballots_input_retrieval_step.py
```diff
@@ -63,12 +65,21 @@ def _get_encryption_devices(
...
+        files = listdir(guardian_keys_dir)
         return [
-            Guardian.from_private_record(g, context.number_of_guardians, context.quorum)
-            for g in guardian_private_records
+            ImportBallotsInputRetrievalStep._get_guardian(guardian_keys_dir, f, context)
+            for f in files
         ]
+
+    @staticmethod
+    def _get_guardian(
```
keithrfung, line 78: This seems like this is actually just a method that should be getting the private records. This could also be pluralized to just get all the records .

```
def load_private_record(
          guardian_dir: str, filename: str
    ) -> PrivateGuardianRecord:
        full_file = join(guardian_dir, filename)
       return from_file(PrivateGuardianRecord, full_file)
```

keithrfung (changes requested):
See comments

keithrfung (approved):

Issue #620, opened 2022-05-02
🚧  Improve Windows installed with gmpy2 preinstalled binary


keithrfung:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Suggestion

We originally had a binary for `gmpy2` to help support Windows installation. With some additions to poetry, this approach seems reproachable for easier installs for windows. 

### Possible Implementation

Use poetry to install a local version of the binary of the 32bit and 64 bit windows.
`poetry add gmpy2 --platform windows --source "./platforms/local-package-of-gmpy2"`

The `pyproject.toml` would resemble something like this at the end.
```yaml
gmpy2 = {version = "2.2.2", platform = "darwin"}
gmpy2 = {version = "2.1.2", platform = "linux"}
gmpy2 = {version = "2.1.2", path="../my/path-64", platform = "windows-64"}
gmpy2 = {version = "2.1.2", path="../my/path-32", platform = "windows-32"}
```

There would have to be a directory of the gmpy2 windows binaries. These can be found here:
https://www.lfd.uci.edu/~gohlke/pythonlibs/

PR #643, opened 2022-05-18
🪟 #620 Fix Makefile on Windows

lprichar:
### Issue

Supports #620 

### Description
Minor fix to the makefile, which didn't run on Windows.  The problem was that the `>` in the embedded python script that determines 32 or 64 bit was being stripped out along with everything after it in any given statement.  It was as if something somewhere detected a cross-site scripting attack and tried to sanitize the input, but did it poorly.  I searched for a while but couldn't find any solutions, so I just removed it, since it wasn't really doing anything anyway.

Incidentally, this PR doesn't address gmpy2, and thus supports but doesn't fix the ticket as written.

### Testing
Run something simple on Windows like `make auto-lint`.  Previously it would give an error about `/5` not being a valid argument or something.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ endif
 
 install-gmp-windows:
 	@echo 🏁 WINDOWS INSTALL
-	@echo 🚨 Ensure pyproject.toml has been modified to include appropriate local gmpy2 package 🚨 
+	@echo 🚨 Nothing todo here. Windows should work via poetry install, without manually installing gmpy2. 🚨 
 # install module with local gmpy2 package
 
 lint:
--- a/Makefile
+++ b/Makefile
@@ -42,12 +42,6 @@ endif
 ifeq ($(OS), Darwin)
 	make install-gmp-mac
 endif
-ifeq ($(OS), Windows)
-	make install-gmp-windows
-endif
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -72,12 +71,6 @@ install-gmp-windows:
 	@echo 🏁 WINDOWS INSTALL
 	@echo 🚨 Ensure pyproject.toml has been modified to include appropriate local gmpy2 package 🚨
```
SteveMaier-IRT, line 72: Since this is working on Windows now with just the poetry install, we can probably get rid of this line

Makefile
```diff
@@ -70,14 +69,8 @@ endif
 
 install-gmp-windows:
 	@echo 🏁 WINDOWS INSTALL
-	@echo 🚨 Ensure pyproject.toml has been modified to include appropriate local gmpy2 package 🚨 
+	@echo 🚨 Nothing todo here. Windows should work via poetry install, without manually installing gmpy2. 🚨
```
keithrfung, line 72: We could likely just remove this whole section.

SteveMaier-IRT (approved):
looks good.

keithrfung (approved):

PR #645, opened 2022-05-25
Story/552 hashed elgamal

### LINKED ISSUE #552: ♻️  Reestablish Secret Sharing using Hashed ElGamal

### Suggestion

Secret sharing used to revolve around Auxiliary Keys. This should be replaced with the new Hashed ElGamal. The messages sent between guardians for the key shares / backups should now use HashedElGamal ciphertext to send messages. 

### Possible Implementation

The guardians sharing a secret (sending a message) will require the public key of the guardian receiving the secret/message. This will then allow the guardians to share the messages securely only to the recipients. This will likely use `HashedElGamalCiphertext`.

### Anything else?

Depends on #551

lprichar:
### Issue

Fixes #552 

### Description
During the key ceremony, guardians share backups aka coordinates with each other that can be used to perform thresshold decryption after the election is complete.  These backups were previously passed and stored in an unencrypted state.  This PR updates the process so that guardians pass and store backups using hashed elgamal (asymetric) encryption.  

In other words when guardian 1 now wants to share a backup to guardian 2, guardian 1 will now encrypt the coordinate using guardian 2's public key.  That coordinate can now only be decrypted using guardian 2's private key.  If guardian 1 was unavailable during the key ceremony, guardian 2 will now first decrypt guardian 1's backup using their private key prior to doing a tally or spoiled ballot decryption.

### Testing
The existing tests thoroughly cover all of the various scenarios around backup sharing, backup validation, and thresshold decryption.

### DIFF
--- a/src/electionguard/decryption.py
+++ b/src/electionguard/decryption.py
@@ -2,6 +2,7 @@ from typing import Dict, List, Optional, Tuple
 from electionguard.chaum_pedersen import ChaumPedersenProof, make_chaum_pedersen
 
 from electionguard.elgamal import ElGamalCiphertext, ElGamalSecretKey
+from electionguard.utils import get_optional
 
 from .ballot import (
     SubmittedBallot,
@@ -29,6 +30,7 @@ from .group import (
     rand_q,
 )
 from .key_ceremony import (
+    CoordinateData,
     ElectionKeyPair,
     ElectionPartialKeyBackup,
     ElectionPublicKey,
@@ -470,7 +472,7 @@ def compensate_decrypt(
     :param missing_guardian_backup: Missing guardians backup
     :param ciphertext: the `ElGamalCiphertext` that will be partially decrypted
     :param extended_base_hash: the extended base hash of the election that
-                                was used to generate t he ElGamal Ciphertext
+                                was used to generate the ElGamal Ciphertext
     :param nonce_seed: an optional value used to generate the `ChaumPedersenProof`
                         if no value is provided, a random number will be used.
     :return: a `Tuple[ElementModP, ChaumPedersenProof]` of the decryption and its proof
@@ -481,11 +483,17 @@ def compensate_decrypt(
     # todo: figure out how to get the guardian's secret_key during compensate_decrypt
     secret_key = ElGamalSecretKey(0)
     encryption_seed = get_backup_seed(
-        guardian_key.owner_id, guardian_key.sequence_order
+        guardian_key.owner_id,
+        missing_guardian_backup.designated_id,
+        guardian_key.sequence_order,
     )
-    partial_secret_key = missing_guardian_backup.encrypted_coordinate.decrypt_to_q(
+    bytes_optional = missing_guardian_backup.encrypted_coordinate.decrypt(
         secret_key, encryption_seed
     )
+    coordinate_data: CoordinateData = CoordinateData.from_bytes(
+        get_optional(bytes_optional)
+    )
+    partial_secret_key = coordinate_data.coordinate
 
     # 𝑀_{𝑖,l} = 𝐴^P𝑖_{l}
     partial_decryption = ciphertext.partial_decrypt(partial_secret_key)
--- a/src/electionguard/decryption.py
+++ b/src/electionguard/decryption.py
@@ -489,6 +489,8 @@ def compensate_decrypt(
     bytes_optional = missing_guardian_backup.encrypted_coordinate.decrypt(
         guardian_keys.key_pair.secret_key, encryption_seed
     )
+    if bytes_optional is None:
+        return None
     coordinate_data: CoordinateData = CoordinateData.from_bytes(
         get_optional(bytes_optional)
     )
@@ -612,7 +614,7 @@ def reconstruct_decryption_contest(
     lagrange_coefficients: Dict[GuardianId, ElementModQ],
 ) -> CiphertextDecryptionContest:
     """
-    Recontruct the missing Decryption Share for a missing guardian
+    Reconstruct the missing Decryption Share for a missing guardian
     from the collection of compensated decryption shares
 
     :param missing_guardian_id: The guardian id for the missing guardian
--- a/src/electionguard/decryption.py
+++ b/src/electionguard/decryption.py
@@ -32,7 +32,7 @@ from .key_ceremony import (
     ElectionKeyPair,
     ElectionPartialKeyBackup,
     ElectionPublicKey,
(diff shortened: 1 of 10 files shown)


src/electionguard/key_ceremony.py
```diff
@@ -179,49 +201,80 @@ def generate_election_key_pair(
...
-        value,
+        sender_guardian_id,
+        receiver_guardian_public_key.owner_id,
+        receiver_guardian_public_key.sequence_order,
+        encrypted_coordinate,
     )
 
 
+def get_backup_seed(receiver_guardian_id: str, sequence_order: int) -> ElementModQ:
+    return hash_elems(receiver_guardian_id, sequence_order)
```
lprichar, line 239: I wasn't sure, is the guardian's id and sequence order sufficient for the backup seed?

src/electionguard/key_ceremony.py
```diff
@@ -163,6 +167,24 @@ class ElectionPartialKeyChallenge:
     coefficient_proofs: List[SchnorrProof]
 
 
+_T = TypeVar("_T", bound="CoordinateData")
+CONTEST_DATA_SIZE: PaddedDataSize = PaddedDataSize.Bytes_512
+
+
+@dataclass
+class CoordinateData:
```
lprichar, line 175: This class seems a bit overkill since it holds a single value, but as discussed with @keithrfung this technique simplifies padding and byte serialization/deserialization.

src/electionguard/decryption.py
```diff
@@ -454,6 +463,7 @@ def compensate_decrypt(
     missing_guardian_backup: ElectionPartialKeyBackup,
     ciphertext: ElGamalCiphertext,
     extended_base_hash: ElementModQ,
+    guardian_keys: ElectionKeyPair,
```
lprichar, line 466: In `compensate_decrypt()` here and also `verify_election_partial_key_backup()` (key_ceremony.py:246), the methods now require a private key of the receiving guardian in order to decrypt the backup.  These signature changes cascade into a lot of other methods and result in some churn in this PR.

src/electionguard/decryption.py
```diff
@@ -463,15 +473,27 @@ def compensate_decrypt(
...
-    partial_secret_key = missing_guardian_backup.coordinate
+    encryption_seed = get_backup_seed(
+        guardian_keys.owner_id,
+        guardian_keys.sequence_order,
+    )
+    bytes_optional = missing_guardian_backup.encrypted_coordinate.decrypt(
+        guardian_keys.key_pair.secret_key, encryption_seed
+    )
+    if bytes_optional is None:
+        return None
```
lprichar, line 492: I do not like propagating this "if null return null pattern", but this PR did not seem like the right place to start fixing it, especially since the source of this problem is in the otherwise untouched hashed elgamal section of code.

src/electionguard/decryption.py
```diff
@@ -81,7 +88,7 @@ def compute_decryption_share(
 
 
 def compute_compensated_decryption_share(
-    guardian_key: ElectionPublicKey,
+    guardian_keys: ElectionKeyPair,
```
keithrfung, line 91: nitpick: `key_pair` should be used. The use of `keys` is confusing and it happens throughout the code. It immediately creates an issue the minute we combine `keys` like for an `ElectionKey`. 

Could be `guardian_key_pair` could also just be `key_pair`. I know it's avoided because of the duplicate `key_pair` within the object.

src/electionguard/decryption.py
keithrfung: The debate can still be made to remove the idea of the backup and simply just pass in the `coordinate`.

I think long term this is the simplest API.
```
def compensate_decrypt(
    missing_guardian_coordinate: ElementModQ,
```

src/electionguard/decryption.py
keithrfung: I agree, but I have no qualms about throwing an error if you wish.

src/electionguard/key_ceremony.py
keithrfung: Not super in love with it, but sometimes simple is better.

src/electionguard/key_ceremony.py
keithrfung: This is up to the specification. I don't think this will likely be enough, but this is a question for @benaloh

tests/unit/test_decrypt_with_shares.py
```diff
@@ -261,7 +261,7 @@ def test_decrypt_ballot_with_missing_guardians(self):
 
         compensated_shares = {
             available_guardian.id: compute_compensated_decryption_share_for_ballot(
-                available_guardian.share_key(),
+                available_guardian._election_keys,
```
keithrfung, line 264: May want to move this line and the similar lines to the `Arrange` since the election_keys are not public and the guardian class would normally be doing it internally. Realistically, this could be done just by manually creating key_pairs instead.

tests/unit/test_decryption.py
```diff
@@ -309,15 +311,15 @@ def test_compute_compensated_selection(self):
 
         # compute compensations shares for the missing guardian
         compensation_0 = compute_compensated_decryption_share_for_selection(
-            available_guardian_1.share_key(),
+            available_guardian_1._election_keys,
```
keithrfung, line 314: See previous comments about not directly accessing private variable in tests.

keithrfung (commented):
Leaving comments. Take a read.

src/electionguard/decryption.py
keithrfung: This is a strong debate to consider going forward with **ElectionGuard** and the future. 

What would be the ideal information to pass into the method?
```py
# Existing
def compensated_decrypt(ciphertext, missing_guardian_backup, extended_base_hash, guardian_keys):
     ...

# Refactored
def decrypt_with_threshold(encrypted_message, key_share, encryption_seed, key_pair):
     ...
```

SteveMaier-IRT (approved):

src/electionguard/decryption.py
```diff
@@ -74,16 +81,16 @@ def compute_decryption_share(
 
     return DecryptionShare(
         tally.object_id,
-        guardian_keys.owner_id,
-        guardian_keys.share().key,
+        key_pair.owner_id,
+        key_pair.share().key,
```
keithrfung, line 85: Is this just to avoid:
`key_pair.key_pair.public_key`

keithrfung (approved):
Minor nitpick. Lgtm.

PR #654, opened 2022-06-07
🐞 Resolve hex string/bytes conversion

### LINKED ISSUE #653: For ElementModQ, q1 == q2 does not imply that hash(q1) == hash(q2)

### Current Behavior

For ElementModQ, q1 == q2 does not imply that hash(q1) == hash(q2)

```
    def testHexToQ(self) -> None :
        s1q = hex_to_q("C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")
        s2q = hex_to_q("0C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")
        self.assertEqual(s1q, s2q)
        self.assertNotEqual(hash_elems(s1q), hash_elems(s2q))
        self.assertNotEqual(s1q.to_hex(), s2q.to_hex())
```

its suprising that leading zeroes are retained, even past the 64 char maximum:

```
    def testHexString(self) -> None :
        self.show("A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")
        self.show("9A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")
        self.show("49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")
        self.show("C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")
        self.show("0C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")
        self.show("00C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")
        self.show("000C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")
        self.show("0000C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")

    def show(self, s1 : str) -> None :
        s1q = hex_to_q(s1)
        print(f" len = {len(s1)} hex = {s1q.to_hex()}")
```

gives:

```
 len = 61 hex = A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206
 len = 62 hex = 9A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206
 len = 63 hex = 49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206
 len = 64 hex = C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206
 len = 65 hex = 0C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206
 len = 66 hex = 00C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206
 len = 67 hex = 000C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206
 len = 68 hex = 0000C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206
```

### Expected Behavior

Should be:

```
    def testHexToQ(self) -> None :
        s1q = hex_to_q("C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")
        s2q = hex_to_q("0C49A1E8053FBA95F6B7CD3F3B30B101CDD595C435A46AECF2872F47F1C601206")
        self.assertEqual(s1q, s2q)
        self.assertEqual(hash_elems(s1q), hash_elems(s2q))
        self.assertEqual(s1q.to_hex(), s2q.to_hex())
```

would recommend that the hex value be exactly 64 characters, with leading zeros truncated or padded as needed.

### Steps To Reproduce

see above

### Environment

```markdown
- OS: all
```

### Anything else?

The same is probably true for ElementModP

In general, the hash function is underspecified.
Porting it to other languages is problematic, since there is no specification other than the python code.
There is a small enough set of argument types to hash_elems() that specifying it should not be too onerous to do.
Then one can test to look for other possible edge cases and make sure there is no ambiguity.

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #653  - Previously 250 in `electionguard`

### Description
- Resolve issue that results from big-integers since they can be treated as byte strings. Leading zeros could be added but result in different results. 
- Add tests to demonstrate

### Testing
The new hash tests replicate the situation displayed in 653

### DIFF
--- a/src/electionguard/big_integer.py
+++ b/src/electionguard/big_integer.py
@@ -9,7 +9,9 @@ from .utils import BYTE_ORDER
 
 def _hex_to_int(input: str) -> int:
     """Given a hex string representing bytes, returns an int."""
-    return int(input, 16)
+    valid_bytes = input[1:] if (len(input) % 2 != 0 and input[0] == "0") else input
+    hex_bytes = bytes.fromhex(valid_bytes)
+    return int.from_bytes(hex_bytes, BYTE_ORDER)
 
 
 def _int_to_hex(input: int) -> str:
@@ -33,8 +35,8 @@ _zero = mpz(0)
 def _convert_to_element(data: Union[int, str]) -> Tuple[str, int]:
     """Convert element to consistent types"""
     if isinstance(data, str):
-        hex = data
         integer = _hex_to_int(data)
+        hex = _int_to_hex(integer)
     else:
         hex = _int_to_hex(data)
         integer = data
--- a/tests/property/test_hash.py
+++ b/tests/property/test_hash.py
@@ -1,12 +1,13 @@
 from typing import List, Optional
 
 from hypothesis import given
+from hypothesis.strategies import integers
 
 
 from tests.base_test_case import BaseTestCase
 
 from electionguard.big_integer import BigInteger
-from electionguard.group import ElementModQ
+from electionguard.group import ElementModP, ElementModQ
 from electionguard.hash import hash_elems
 from electionguard_tools.strategies.group import elements_mod_p, elements_mod_q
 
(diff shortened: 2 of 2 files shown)


src/electionguard/big_integer.py
```diff
@@ -9,7 +9,9 @@
 
 def _hex_to_int(input: str) -> int:
     """Given a hex string representing bytes, returns an int."""
-    return int(input, 16)
+    valid_bytes = input[1:] if (len(input) % 2 != 0 and input[0] == "0") else input
+    hex_bytes = bytes.fromhex(valid_bytes)
```
keithrfung, line 13: This will throw a `ValueError` for strings that are invalid hexadecimal giving a meaningful error. The previous line is a safeguard for accidental odd leading zeroes exclusively. It could be argued these are not legitimate hex at all and that could be removed.

keithrfung:
CC @JohnLCaron 

This isn't as ideal as adding full padding for zeroes but as C++ already handles this in a similar fashion. This ensures hex passed in and the ones out match. I think 2.0 will probably bring shifts towards a more `byte` focused mentality.

tests/property/test_hash.py
```diff
@@ -28,6 +30,27 @@ def test_basic_hash_properties(self, a: ElementModQ, b: ElementModQ):
...
+    @given(elements_mod_p())
+    def test_hash_of_big_integer(self, input: ElementModQ) -> None:
+        """Test hashing of larger integers such as element mod p"""
+
+        # Arrange.
+        input_hash = hash_elems(input)
+        invalid_hex = "0" + input.to_hex()
+        leading_zeroes = "00" + input.to_hex()
+
+        # Act.
```
lprichar, line 42: This looks like two tests in one.  I don't know if this is idiomatic for Python, but personally, I prefer multiple tests with a single act clause each. That way when a test fails you can read the name of the test and know what went wrong.

lprichar (approved):
Nice fix! So glad to see this wasn't as large as we feared it might be.

tests/property/test_hash.py
```diff
@@ -28,6 +30,27 @@ def test_basic_hash_properties(self, a: ElementModQ, b: ElementModQ):
         if ha != hb:
             self.assertNotEqual(a, b)
 
+    @given(elements_mod_p())
+    def test_hash_of_big_integer(self, input: ElementModQ) -> None:
+        """Test hashing of larger integers such as element mod p"""
```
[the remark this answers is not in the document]
keithrfung: Good catch. Resolved.

Issue #659, opened 2022-06-21
🖨️ Print contest and candidate names when printing the tally in the CLI


lprichar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Suggestion

Print contest and candidate names when printing the tally in the CLI.  For candidate names assume english, but make it a parameter.

PR #660, opened 2022-06-21
🖨️ #29 print names

lprichar:

### Issue

Addresses #659

### Description
Now when you import ballots and do a tally the CLI will display contest names and candidate names rather than the id's.

### Testing
New tests cover most of the core functionality, however run an import to be sure.

### DIFF
--- a/src/electionguard/manifest.py
+++ b/src/electionguard/manifest.py
@@ -807,7 +807,7 @@ class Manifest(CryptoHashable):
 
     def _get_candidate_name(self, candidate: Candidate, lang: str) -> str:
         if candidate.is_write_in:
-            return "write-in"
+            return "Write-In"
         query = (t.value for t in candidate.name.text if t.language == lang)
         name = next(query, None)
         name = candidate.object_id if name is None else name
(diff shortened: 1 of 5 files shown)


SteveMaier-IRT (approved):
Looks good.

src/electionguard/manifest.py
```diff
@@ -805,6 +805,53 @@ def is_valid(self) -> bool:
...
+        """
+        Retrieves a dictionary whose keys are all selection id's and whose values are
+        those selection's candidate names in the supplied language if available
+        """
+        candidates = self._get_candidate_names(lang)
+        selections = self._get_selections_with_candidate_id()
+        self._replace_candidate_ids_with_names(selections, candidates)
+        return selections
+
+    def get_contest_names(self) -> Dict[str, str]:
```
keithrfung, line 847: These are pretty specific and not a requirement of something that should be a `dataclass`. I would move these to functional methods or a wrapping class both of which could exist in `electionguard_tools/helpers` or the `cli` directly.

`get_contest_names(manifest: Manifest)`

or 

`class PrintableManifest (Manifest)`

keithrfung (commented):
Just commenting these, but I don't think these methods should be where they are.

Issue #651, opened 2022-06-06
🐞 HashedElGamalCiphertext is buggy, fails inconsistently


danwallach:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

In my attempts to understand HashedElGamal, I've found problems. I wrote this unit test, while setting the environment variable to ensure I was using the 4096-bit production group.

```python
    @given(elgamal_keypairs(), binary(min_size=10, max_size=400))
    def test_hashed_elgamal_properties(self, kp: ElGamalKeyPair, plaintext: bytes) -> None:
        hmac_nonce = rand_range_q(2)
        hmac_seed = rand_range_q(2)
        padded_plaintext = _add_padding(plaintext, PaddedDataSize.Bytes_512)
        hmac = hashed_elgamal_encrypt(padded_plaintext, hmac_nonce, kp.public_key, hmac_seed)
        decryption_bytes_padded = hmac.decrypt(kp.secret_key, hmac_seed)
        self.assertIsNotNone(decryption_bytes_padded)

        decryption_bytes = _remove_padding(decryption_bytes_padded, PaddedDataSize.Bytes_512)
        self.assertEqual(plaintext, decryption_bytes)
```

This test passes often. But it sometimes fails the HMAC check.

```
-------------------------------- Captured stdout call --------------------------------
[48413:2022-06-06 16:23:24,477]:ERROR:elgamal.py.decrypt:#L143: MAC verification failed in decryption.
--------------------------------- Captured log call ----------------------------------
ERROR    electionguard:logs.py:87 elgamal.py.decrypt:#L143: MAC verification failed in decryption.
------------------------------------- Hypothesis -------------------------------------
Falsifying example: test_hashed_elgamal_properties(
[...]

PR #666, opened 2022-06-23
🔥 Hot Fix Ceremony for PR's

keithrfung:

### Issue
*Link your PR to an issue*

Reference to #651 

### Description
This does not resolve the issue. This side steps the ceremony during tests until this is resolved. The thresholding steps (aka steps 2 and 3) utilize the HashedElgamal which is failing in certain cases. This is issue #651 and needs to be resolved.

### Testing
This doesn't remove the check from test_end_to_end_election so this should simply reduce the issues. It aims to reduce not eliminate the presence of the issue. This should be an overall improvement for speed on tests that don't require full ceremony.

### DIFF
--- a/src/electionguard/key_ceremony_mediator.py
+++ b/src/electionguard/key_ceremony_mediator.py
@@ -171,7 +171,7 @@ class KeyCeremonyMediator:
         Publish joint election key from the public keys of all guardians
         :return: Joint key for election
         """
-        if not self.all_backups_verified():
+        if not self.all_guardians_announced():
             return None
 
         return combine_election_public_keys(list(self._election_public_keys.values()))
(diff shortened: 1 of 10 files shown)


src/electionguard_tools/helpers/key_ceremony_orchestrator.py
```diff
@@ -21,8 +21,19 @@ def create_guardians(ceremony_details: CeremonyDetails) -> List[Guardian]:
         ]
 
     @staticmethod
-    def perform_full_ceremony(guardians: List[Guardian], mediator: KeyCeremonyMediator):
-        """Perform full key ceremony so joint election key is ready for publish"""
+    def perform_ceremony(guardians: List[Guardian], mediator: KeyCeremonyMediator):
```
keithrfung, line 24: The key piece.

src/electionguard/key_ceremony_mediator.py
```diff
@@ -171,7 +171,7 @@ def publish_joint_key(self) -> Optional[ElectionJointKey]:
         Publish joint election key from the public keys of all guardians
         :return: Joint key for election
         """
-        if not self.all_backups_verified():
+        if not self.all_guardians_announced():
```
keithrfung, line 174: Part 2. 

This can stay long term.

PR #670, opened 2022-06-27
🐞 Refactor - CLI models should use dataclass instead of init

keithrfung:

### Issue
*Link your PR to an issue*

Fixes #___

### Description
Dataclass autogenerates the exact same init since there is no specialization on these files. This shortens the classes down and simplifies constructors.

https://docs.python.org/3/library/dataclasses.html

### Testing
CLI should function exactly the same. This is a net same code.


lprichar (approved):
👍

keithrfung:
This needs the mypy one to pass. The reason that is failing is actually related to the types.

keithrfung:
Blocked by #678

PR #679, opened 2022-06-29
🐛 #651 hashed elgamal inconsistent results

### LINKED ISSUE #651: 🐞 HashedElGamalCiphertext is buggy, fails inconsistently

lprichar:
### Issue

Fixes #651 

### Description
First of all, big thanks to @cryptosidh for assistance diagnosing this.  The problem was that certain nonce/seed/data combinations would produce an encrypted value that in bytes started with a zero.  When `hashed_elgamal_encrypt()` returns that data it calls bytes_to_hex() which truncates leading zero's. That produces 255 bytes of encrypted data instead of 256 bytes and that doesn't work with the XOR based encryption/decryption.  The solution is to pad the encrypted data in `decrypt()` if there are fewer than 256 bytes.

### Testing
There are two new unit tests for both the positive and negitive cases of leading zero's in the ciphertext.


src/electionguard/elgamal.py
```diff
@@ -24,6 +24,7 @@
 ElGamalPublicKey = ElementModP
 
 _BLOCK_SIZE = 32
+_ELGAMAL_DATA_SIZE = 512
```
keithrfung, line 27: I feel like this can and should use the Padded Data Size that already exists. These should be using the same number but more importantly ARE the same number. They are just set at different points. 

https://github.com/microsoft/electionguard-python/blob/5b73f75cc54f63837bf7cd6328adff208e6dca62/src/electionguard/serialize.py#L51

tests/unit/test_elgamal.py
```diff
@@ -0,0 +1,63 @@
+from electionguard import ElGamalKeyPair
+from electionguard.elgamal import (
+    ElGamalPublicKey,
+    ElGamalSecretKey,
+    hashed_elgamal_encrypt,
+)
+from electionguard.group import ElementModQ
+from electionguard.serialize import PaddedDataSize, _add_padding, _remove_padding
```
keithrfung, line 8: If we are using the `_add_padding` and `_remove_padding` like this. It should be denoted as a public function.

keithrfung (changes requested):

tests/unit/test_elgamal.py
```diff
@@ -0,0 +1,63 @@
...
+        self.do_hashed_elgamal(kp, plaintext, hmac_nonce, hmac_seed)
+
+    def do_hashed_elgamal(
+        self,
+        kp: ElGamalKeyPair,
+        plaintext: bytes,
+        hmac_nonce: ElementModQ,
+        hmac_seed: ElementModQ,
+    ) -> None:
+        padded_plaintext = _add_padding(plaintext, PaddedDataSize.Bytes_512)
```
keithrfung, line 53: This is a good example of what I was discussing before with the use of PaddedDataSize here. 

1. Pad `plaintext` to `512` bytes
2. `hashed_elgamal_decrypt` while adding one more bit of padding to make sure it's `512`

keithrfung: We're only working in 512 with this whole encryption so we should have it made much more obvious the whole way through.

keithrfung: https://github.com/microsoft/electionguard-python/blob/5b73f75cc54f63837bf7cd6328adff208e6dca62/src/electionguard/serialize.py#L51-L55

perhaps this
```py
 class DataSize(IntEnum): 
     """Define the sizes for data.""" 
  
     Bytes_512 = 512
     
def get_padded_data_size(data_size: DataSize): 
      return DataSize - PAD_INDICATOR_SIZE
```

keithrfung: This also works but it'll expose the items for modification while it's running.

```py
 DATA_SIZE = 512
 PAD_INDICATOR_SIZE = 2
 PADDED_DATA_SIZE = DATA_SIZE - PAD_INDICATOR_SIZE
 ```

Comment (keithrfung) [reply tests/unit/test_elgamal.py:53]: This line is probably relevant in making decisions
https://github.com/microsoft/electionguard-python/blob/5b73f75cc54f63837bf7cd6328adff208e6dca62/src/electionguard/serialize.py#L91

File: src/electionguard/serialize.py [hunk 4]
```diff
@@ -48,12 +48,6 @@
 )
 
 
-class PaddedDataSize(IntEnum):
```
keithrfung, line 51: The add and remove padding functions should be next to this.

src/electionguard/padded_data_size.py
```diff
@@ -0,0 +1,10 @@
+from enum import IntEnum
+
+PAD_INDICATOR_SIZE = 2
+
+
+class PaddedDataSize(IntEnum):
+    """Define the sizes for padded data."""
+
+    DataSize = 512
+    Bytes_512 = DataSize - PAD_INDICATOR_SIZE
```
keithrfung, line 10: This doesn't resolve the issue. The DataSize is NOT a PaddedDataSize. See my previous code snippets.

lprichar: I took another stab at it.  lmk if the latest code looks good, else please provide a more specific explanation of what you'd like to see changed.

src/electionguard/elgamal.py
lprichar: Steve is correct, but I can see how they're tightly related and there should probably be an underlying 512 constant, so that's what it does currently

src/electionguard/byte_padding.py
```diff
@@ -0,0 +1,54 @@
...
+
+_PAD_BYTE = b"\x00"
+_BYTE_ORDER = "big"
+_PAD_INDICATOR_SIZE = 2
+
+
+class DataSize(IntEnum):
+    """Define the sizes for data."""
+
+    Bytes_512 = 512
```
keithrfung, line 12: There is now a single data size used. The padding methods are all together and everything defaults to the only current size we support.

keithrfung (approved):

PR #700, opened 2022-07-12
⚡ Submit (Cast/Spoil) Encrypted Ballots in CLI

### LINKED ISSUE #585: ⚡️  Submit (Cast/Spoil) Encrypted Ballots in CLI

### Suggestion

This would bypass the step of encrypting the votes (which the CLI does today).  The system should support both casting and spoiling ballots.  Initially this should be based on a directory e.g. `/import/cast` contains ballots to cast and `/import/spoil` contains ballots to spoil.

### Possible Implementation

_No response_

### Anything else?

_No response_

shreyasminocha:

### Issue

Fixes #585 

### Description
```sh
eg submit-ballots /tmp/ciphertext-ballots/cast /tmp/ciphertext-ballots/spoil /tmp/submitted-ballots
eg submit-ballots /tmp/ciphertext-ballots/cast /tmp/submitted-ballots # no spoilt ballots
```

The spoil ballots directory is currently optional, but I can make it mandatory if that's more appropriate. 

Also, I noticed that the deprecated `sample_generator.py` prefixes its output filenames with strings from `electionguard_tools.helpers.export`. Let me know if we want those prefixes here too.


SteveMaier-IRT (approved):
looks good

PR #704, opened 2022-07-14
#702 Make `mark-ballots` selections match `number_elected`

### LINKED ISSUE #702: 🐞 In the CLI, the ballot generation has an issue with the pick 2 or 3 ballots

### Current Behavior

The current plain text ballot generation will make ballots for a manifest, if there are contests with the number_elected set to 2 or 3, then there might not be 2 or 3 selections picked in the ballot generated.  The generation process needs to be able to vote for the number that is in the manifest.  

### Expected Behavior

Have the number selected in the plain text ballot be the same as the number_elected for most of not all of the ballots.

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

shreyasminocha:

### Issue

Fixes #702

### Description
`eg mark-ballots` doesn't undervote or null vote anymore.

### Testing
Besides verifying that this fixes #702, double-check that this doesn't accidentally reduce the set of ballots we test against.

### DIFF
--- a/src/electionguard_cli/cli_steps/mark_ballots_step.py
+++ b/src/electionguard_cli/cli_steps/mark_ballots_step.py
@@ -18,6 +18,10 @@ class MarkBallotsStep(CliStepBase):
         internal_manifest = build_election_results.internal_manifest
         ballot_factory = BallotFactory()
         plaintext_ballots = ballot_factory.generate_fake_plaintext_ballots_for_election(
-            internal_manifest, num_ballots, ballot_style_id
+            internal_manifest,
+            num_ballots,
+            ballot_style_id,
+            allow_null_votes=False,
+            allow_under_votes=False,
         )
         return MarkResults(plaintext_ballots)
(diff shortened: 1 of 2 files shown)


shreyasminocha:
While the changes,are ready one of every ten-or-so test runs fails with issues related to a zero nonce. This seems to happen only with the tests in `property/test_decryption_mediator.py` and the exception arises either in `elgamal.py` (where there's an explicit check for zero nonces) or in `group.py` (where negating zero results in a value that's too large for the group). I suspect the fix for that is out-of-scope for this PR, but since it's these changes to the tests that cause the failures, I want to try to track it down before we merge.

shreyasminocha:
Nevermind, these failures occur on the main branch too, although they seem much rarer than I initially estimated; I'll create a separate issue to track those. Meanwhile, this is ready for review.

SteveMaier-IRT (approved):
looks good.

PR #705, opened 2022-07-14
Fix sample data CI failure

lprichar:
### Issue

na

### Description
Sample data was previously being double-zipped and then in this repo double unzipped.  That was fixed recently and that caused CI to fail.  This fixes it by removing the double-unzipping.

### Testing
CI should pass.


SteveMaier-IRT (approved):
looks good.

PR #712, opened 2022-07-21
#691 Make Joint Key, Complete Key Ceremony in GUI

### LINKED ISSUE #691: ✨ Admin Publish Joint Key, Generate Context (#6e, #7)

### Suggestion

As an admin when I receive backup verifications from all guardians then the system should receive the verification, the publish joint key, generate the context, and display the key ceremony complete page.

### Possible Implementation

_No response_

### Anything else?

_No response_

lprichar:
### Issue

Fixes #691 

### Description
After the guardians verify backups the admin now receives the verifications and publishes the joint key.  This PR additionally:

- Adds generalized exception handling to the key ceremony process
- Fixes issues with sluggish UI based on how the long polling was working
- Updates the UI to be prettier during the process and once completed
- Records the date/time when the key ceremony was completed in the DB

### Testing
1. Have an admin create a key ceremony
2. Have guardians join the key ceremony

The process should complete successfully and show a green checkmark.  It should look like this:

key-ceremony-10


src/electionguard_gui/components/key_ceremony_details_component.py
```diff
@@ -81,27 +76,35 @@ def watch_key_ceremony(self, key_ceremony_id: str) -> None:
...
+                key_ceremony
+            )
+            if state != new_state:
+                self.log.debug(f"state changed from {state} to {new_state}")
+            key_ceremony.status = get_key_ceremony_status(new_state)
+            result = key_ceremony.to_dict()
+            # pylint: disable=no-member
+            eel.refresh_key_ceremony(eel_success(result))
+        # pylint: disable=broad-except
+        except Exception as e:
```
lprichar, line 104: Apparently github marks content as changed if you simply indent.  The try/catch for generalized exception handling was the only thing actually added here.

src/electionguard_gui/services/key_ceremony_service.py
```diff
@@ -68,7 +71,7 @@ def watch_key_ceremonies(
                 # the tailable cursor times out after a few seconds and fires a StopIteration exception,
                 # so we need to catch it and restart watching. The sleep is required by eel to allow
                 # it to respond to events such as the very important stop_watching event.
-                eel.sleep(0.2)
+                eel.sleep(0.8)
```
lprichar, line 74: spending more time in the eel sleep makes the UI substantially more responsive, and the .2s long polling is still enough to give super-fast feedback cycles during the multiple back-and-forth's of the key ceremony.

src/electionguard_gui/web/components/shared/key-ceremony-details-component.js
```diff
@@ -6,17 +6,23 @@ export default {
...
     },
-    refresh_key_ceremony: function (keyCeremony) {
-      console.log("key ceremony refreshed", keyCeremony);
-      this.keyCeremony = keyCeremony;
+    refresh_key_ceremony: function (eelMessage) {
+      console.log("key ceremony refreshed", eelMessage);
+      if (eelMessage.success) {
+        this.keyCeremony = eelMessage.result;
+      } else {
+        console.error(eelMessage.message);
```
lprichar, line 22: In a traditional _web app_ returning server-side error messages to the client-side could be a security concern, but I believe it's fine in this environment because the user already has full access to the machine the _desktop app_ is running on.  I'll fix if anyone disagrees.

src/electionguard_gui/web/components/shared/key-ceremony-details-component.js
```diff
@@ -6,17 +6,23 @@ export default {
...
+        this.error = true;
```
SteveMaier-IRT, line 23: Do we need to set the error to false in the if above this?  since it is in a refresh call, it could be that it is an error first and then it is successful but the error flag will still be true.

PR #713, opened 2022-07-21
Draft Solution CLI Print Methods

Alopezjr2002:
Print_utils.py includes methods that will handles all click.echo references. Instead of having click.echo through the code base it can be found only once inside of the file created


SteveMaier-IRT:
@Alopezjr2002 please make sure to sign the CLA.

PR #714, opened 2022-07-21
Draft Solution CLI Print Methods Issue

CodingForFun-coder:
The new file that was added was print_utils.py which includes all click.echo references that were in electionguard_cli. All files that had contained click.echo within electionguard_cli now has methods from prnt_utils.py that contain click.echo references.


SteveMaier-IRT:
@CodingForFun-coder please make sure that you sign the CLA.

src/electionguard_cli/cli_steps/cli_step_base.py
```diff
@@ -15,17 +15,17 @@ class CliStepBase:
     VERIFICATION_URL_NAME = "verification_url"
 
     def print_header(self, s: str) -> None:
-        click.echo("")
-        click.secho(f"{'-'*40}", fg=self.header_color)
-        click.secho(s, fg=self.header_color)
-        click.secho(f"{'-'*40}", fg=self.header_color)
+        Echo("")
+        Secho(f"{'-'*40}", fg=self.header_color)
```
shivbijlani, line 19: Move header_color and other local members to print utils. 
CliStepBase should not care about the colors of printing to the command line (separation of concerns). It should call print header method or print warning method, and the implementation details of print module should deal with colors.

shivbijlani: cc @CodingForFun-coder , @jknight777

PR #715, opened 2022-07-21
Draft CLI Print Method

Alopezjr2002:
Print_utils.py includes methods that will handles all click.echo references. Instead of having click.echo through the code base it can be found only once inside of the file created


src/electionguard_cli/print_utils.py
```diff
@@ -1,19 +1,26 @@
...
-def print_message(text: str, underlined=False):
+def print_message_utils(text: str, underlined=False):
     click.secho(f"{text}", fg="white", underline=underlined)
 
-def print_warning(text: str, warning_color="bright_red") -> None:
+def print_warning_utils(text: str, warning_color="bright_red") -> None:
     click.secho(f"WARNING: {text}", fg=warning_color, bold=True)
 
-def print_error(text: str, error_color="bright_red"):
+def print_error_utils(text: str, error_color="bright_red"):
```
shivbijlani, line 16: Whats the reason for adding _utils at the end of each function?

Alopezjr2002: inside of cli_step_base print_header was the name of the method that was already existing and just to distinguish which print_header belonged where. If I didnt add _utils it would have confused me as to which method I was using

shivbijlani: I understand, but those methods are in different files. to disambiguate, try using the "import as" functionality in python so that print_header in the base calls print_utils.print_header in the new print module.

PR #717, opened 2022-07-23
Draft CLI Print Method

CodingForFun-coder:
The new file that was added was print_utils.py which includes all click.echo references that were in electionguard_cli. All files that had contained click.echo within electionguard_cli now has methods from prnt_utils.py that contain click.echo references.


src/electionguard_cli/cli_steps/cli_step_base.py
```diff
@@ -1,5 +1,5 @@
 from typing import Any, Optional
-import click
+from print_utils import *
```
shivbijlani, line 2: In the final PR, remove click.echo's from this file. See Antonio's pr

tests/property/test_verify.py
```diff
@@ -46,7 +46,7 @@ class TestVerify(BaseTestCase):
     """Test ballot verification"""
 
     @settings(
-        deadline=timedelta(milliseconds=2000),
+        deadline=timedelta(milliseconds=5000),
```
shivbijlani, line 49: These changes be showing because your fork is behind master? Click the sync button on your fork to get the latest from master

PR #720, opened 2022-07-25
#694 Upload Ballots

### LINKED ISSUE #694: ✨ Upload Ballots in GUI

### Suggestion

As an admin I want to select an election and a folder full of ballots to upload so that I can subsequently decrypt them

### Possible Implementation

_No response_

### Anything else?

_No response_

lprichar:
### Issue
Fixes #694 

### Description
Provides the ability to upload ballots to an election.

upload-ballots

### Testing
1. Create an election
2. Select the new upload icon in the upper right

image

3. Select a device file and ballot folder

image

4. Click upload

Expected: a success screen, see gif in the Description section

### DIFF
--- a/src/electionguard_gui/services/ballot_upload_service.py
+++ b/src/electionguard_gui/services/ballot_upload_service.py
@@ -1,0 +1,55 @@
+from datetime import datetime
+from bson import ObjectId
+from pymongo.database import Database
+from electionguard_gui.services.eel_log_service import EelLogService
+from electionguard_gui.services.service_base import ServiceBase
+from electionguard_gui.services.authorization_service import AuthorizationService
+
+
+class BallotUploadService(ServiceBase):
+    """Responsible for functionality related to ballot uploads"""
+
+    _log: EelLogService
+    _auth_service: AuthorizationService
+
+    def __init__(
+        self, log_service: EelLogService, auth_service: AuthorizationService
+    ) -> None:
+        self._log = log_service
+        self._auth_service = auth_service
+
+    def create(
+        self,
+        db: Database,
+        election_id: str,
+        device_file_name: str,
+        device_file_contents: str,
+    ) -> str:
+        election = {
+            "election_id": election_id,
+            "device_file_name": device_file_name,
+            "device_file_contents": device_file_contents,
+            "ballots": [],
+            "created_by": self._auth_service.get_user_id(),
+            "created_at": datetime.utcnow(),
+        }
+        self._log.trace(f"inserting ballot upload for: {election_id}")
+        inserted_id = db.ballot_uploads.insert_one(election).inserted_id
+        return str(inserted_id)
+
+    def add_ballot(
+        self,
+        db: Database,
+        ballot_upload_id: str,
+        file_name: str,
+        file_contents: str,
+    ) -> None:
+        self._log.trace(f"adding ballot {file_name} to {ballot_upload_id}")
+        db.ballot_uploads.update_one(
+            {"_id": ObjectId(ballot_upload_id)},
+            {
+                "$push": {
+                    "ballots": {"file_name": file_name, "file_contents": file_contents}
+                }
+            },
+        )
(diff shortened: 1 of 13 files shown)


src/electionguard_gui/services/ballot_upload_service.py
```diff
@@ -0,0 +1,55 @@
...
+        election = {
+            "election_id": election_id,
+            "device_file_name": device_file_name,
+            "device_file_contents": device_file_contents,
+            "ballots": [],
+            "created_by": self._auth_service.get_user_id(),
+            "created_at": datetime.utcnow(),
+        }
+        self._log.trace(f"inserting ballot upload for: {election_id}")
+        inserted_id = db.ballot_uploads.insert_one(election).inserted_id
```
lprichar, line 37: One thing I struggle with in NoSQL databases is when to use a new collection vs how much stuff to stuff into an existing record of an existing collection.  This felt like a new collection with a "foreign key" kind of situation, otherwise the election record would just get gigantic with all those ballots in it.

src/electionguard_gui/web/components/admin/upload-ballots-component.js
```diff
@@ -0,0 +1,122 @@
...
+      if (!result.success) {
+        throw new Error(result.message);
+      }
+      this.ballotsProcessed++;
+      return result.result;
+    },
+    async uploadBallotFiles(uploadId, ballotFiles) {
+      for (let i = 0; i < ballotFiles.length; i++) {
+        const ballotFile = ballotFiles[i];
+        const ballotContents = await ballotFile.text();
```
lprichar, line 60: Reading in all the ballot files in JavaScript and sending the contents over web sockets to store in Mongo may not end up being the best approach since web sockets theoretically aren't as reliable as HTTP, but it seems to work well for me and it's all local traffic to the device.  If it ends up being a problem, we'll have to think pretty hard to find any better options.

src/electionguard_gui/web/components/admin/view-election-component.js
```diff
@@ -28,54 +34,69 @@ export default {
...
       <div class="container">
-        <div class="row">
-          <div class="col col-10">
+        <div class="row mb-4">
+          <div class="col">
             <h1>{{election.election_name}}</h1>
           </div>
-          <div class="col col-2 text-end">
-            <a :href="getEncryptionPackageUrl()" class="btn btn-primary">
+          <div class="col col-xs-3 text-end">
```
lprichar, line 41: I don't love this UI for downloading the encryption package and uploading ballots.  Open to suggestions.

src/electionguard_gui/web/components/admin/create-election-component.js
```diff
@@ -15,35 +15,29 @@ export default {
     Spinner,
   },
   methods: {
-    createElection() {
+    async createElection() {
```
lprichar, line 18: This was a somewhat unrelated refactoring that makes me happy.  When I wrote the create election component I wasn't as familiar with how to read files and call eel's methods with the async/await pattern.  I love this refactoring because it shows how much more readable things are with async/await.  It makes me very happy that Anders Hejlsberg introduced the pattern to the world in C# 5 in 2012 and it was so popular that JavaScript stole it five years later in 2017.

src/electionguard_gui/web/components/admin/upload-ballots-component.js
SteveMaier-IRT: Since the ballot uploading is from the admin device and that is the one with the database on it, this should not be an issue at all.

src/electionguard_gui/services/ballot_upload_service.py
SteveMaier-IRT: The other option is a ballot record type inside of the election record collection.  This way you can query for just one type or the other and it is all inside of a single collection.

SteveMaier-IRT (approved):
Looks good.

PR #721, opened 2022-07-26
#695 Create Decryption

### LINKED ISSUE #695: ✨ Start Decryption in GUI

### Suggestion

As an admin I want to start a decryption for an election that will decrypt records so that I can view tally and publish an election record

### Possible Implementation

_No response_

### Anything else?

_No response_

lprichar:
### Issue

Fixes #695 

### Description
Adds the ability to create decryptions for an election and view those decryptions.  It does not yet include the ability for guardians to join a decryption.

### Testing
1. Create an election
2. Click the "Create Decryption" button at the top right

Expected: It should navigate to a create decryption page and the name should be auto-filled in with a suggested value

image

3. Click create

Expected: It should navigate to a view decryption page like this:

image


src/electionguard_gui/components/component_base.py
```diff
@@ -22,3 +25,8 @@ def init(
     def expose(self) -> None:
         """Override to expose the component's methods to JavaScript. This technique hides the
         fact that method names exposed must be globally unique."""
+
+    def handle_error(self, error: Exception) -> dict[str, Any]:
```
lprichar, line 29: This is an improvement, but what I really wanted was a `run_safe` type method that would take a lambda, create a database, and then call the lambda with the db in a try/catch block.  Unfortunately, Python [doesn't support](https://www.artima.com/weblogs/viewpost.jsp?thread=147358) multi-line lambdas, so I couldn't get this to work.  I may come back to it.  I would also welcome suggestions.

src/electionguard_gui/components/upload_ballots_component.py
```diff
@@ -25,23 +25,39 @@ def expose(self) -> None:
...
             ballot_upload_id = self._ballot_upload_service.create(
-                db, election_id, device_file_name, device_file_contents
+                db,
+                election_id,
+                device_file_name,
+                device_file_contents,
+                ballot_count,
+                created_at,
+            )
+            self._election_service.append_ballot_upload(
```
lprichar, line 49: Duplicated ballot info into the election so it could easily get displayed on the election page

src/electionguard_gui/services/election_service.py
```diff
@@ -45,6 +46,7 @@ def create_election(
             "key_ceremony_id": key_ceremony.id,
             "guardians": context.number_of_guardians,
             "quorum": context.quorum,
+            "election_url": election_url,
```
lprichar, line 49: Forgot to save the url in a prior PR.  It was being put into the context, but now it gets displayed in the UI too, which is kinda nice.

src/electionguard_gui/web/components/admin/create-election-component.js
```diff
@@ -1,5 +1,5 @@
 import Spinner from "../shared/spinner-component.js";
-import RouterService from "/services/router-service.js";
+import RouterService from "../../services/router-service.js";
```
lprichar, line 2: TIL VS Code only gives you code completion when you use relative references

src/electionguard_gui/services/decryption_service.py
```diff
@@ -0,0 +1,57 @@
...
+            "created_by": self._auth_service.get_user_id(),
+            "created_at": datetime.utcnow(),
+        }
+        self._log.trace(f"inserting decryption for: {election.id}")
+        inserted_id = db.decryptions.insert_one(decryption).inserted_id
+        return str(inserted_id)
+
+    def name_exists(self, db: Database, name: str) -> Any:
+        self._log.trace(f"getting decryption by name: {name}")
+        decryption = db.decryptions.find_one({"name": name})
```
SteveMaier-IRT, line 46: Should this be searching for the election_name or the decryptrion_name?

SteveMaier-IRT (changes requested):
One possible issue.  Other than that, looks good.

src/electionguard_gui/services/decryption_service.py
lprichar: Great catch!

SteveMaier-IRT (approved):
looks good.

PR #724, opened 2022-07-28
#722 Join Decryption

### LINKED ISSUE #722: ✨ Join Decryption in GUI

### Suggestion

As a guardian I want to join a decryption.  Once all guardians have joined the decryption should take place.

### Possible Implementation

_No response_

### Anything else?

_No response_

lprichar:
### Issue

Fixes #722 

### Description
Guardians can now join decryptions.  When they do they use their private key to create a partial decryption which is stored in the database.  Once all guardians have provided their partial decryptions the admin automatically decrypts and stores the result back into the database.  Neither the tally nor the spoiled ballots are displayed yet, that will be implemented in a subsequent PR.

### Testing
1. Create an election
2. Create a decryption
3. Have guardians log in, they should see a new decryption like:

image

4. Have guardian click the decryption, their screen should look like this:

image

5. Have all guardians click the join button.  

After the 1st guardian clicks the admin's screen should look like this:

image

After all guardians have joined the admin screen should look like this:

image


src/electionguard_db/mongo-init.js
```diff
@@ -3,5 +3,5 @@ db.createCollection("key_ceremonies");
 db.createCollection("elections");
 db.createCollection("ballot_uploads");
 db.createCollection("decryptions");
-db.createCollection("key_ceremony_deltas", { capped: true, size: 100000 });
-db.key_ceremony_deltas.insert({ type: "init" });
+db.createCollection("db_deltas", { capped: true, size: 100000 });
```
lprichar, line 6: this was reworked to be collection agnostic

src/electionguard_gui/components/guardian_home_component.py
```diff
@@ -0,0 +1,67 @@
...
+        ]
+        return eel_success(js_key_ceremonies)
+
+    def watch_db_collections(self) -> None:
+        self._log.debug("Watching database")
+        db = self._db_service.get_db()
+        self._db_watcher_service.watch_database(db, None, notify_ui_db_changed)
+        self._log.debug("exited watching key_ceremonies")
+
+    def stop_watching_key_ceremonies(self) -> None:
```
lprichar, line 57: bad name, will fix

src/electionguard_gui/components/view_decryption_component.py
```diff
@@ -1,31 +1,100 @@
...
+    DecryptionService,
+    DbWatcherService,
+)
+from electionguard_gui.services.decryption_stages import (
+    DecryptionS1JoinService,
+    DecryptionS2AnnounceService,
+)
 
 
 class ViewDecryptionComponent(ComponentBase):
```
lprichar, line 19: In a subsequent PR I'll probably split this out into one that's specific to admin and one that's specific to guardian

src/electionguard_gui/services/db_watcher_service.py
```diff
@@ -0,0 +1,69 @@
...
+        if self.watching_database.is_set():
+            self.stop_watching()
+
+        # set a semaphore to indicate that we are watching the database
+        self.watching_database.set()
+        while self.watching_database.is_set() and cursor.alive:
+            try:
+                # block for up to a few seconds until someone adds a new delta
+                delta = cursor.next()
+                collection = delta["collection"]
```
lprichar, line 56: This code is almost identical to the `watch_key_ceremonies` method that was in the key ceremony service.  The only difference is this bit that tracks the collection that changed, which the guardian home page uses to determine which section to update.  This change was required because this code is necessarily single-threaded and so a given machine can only be performing one watching operation at a time.  If both the key cermeony and the decryption sections were watching it would violate that.

src/electionguard_gui/services/decryption_stages/decryption_s1_join_service.py
```diff
@@ -0,0 +1,45 @@
...
+from electionguard.ballot import BallotBoxState
+
+from electionguard_gui.models.decryption_dto import DecryptionDto
+from electionguard_gui.services.decryption_stages.decryption_stage_base import (
+    DecryptionStageBase,
+    get_tally,
+)
+
+
+class DecryptionS1JoinService(DecryptionStageBase):
```
lprichar, line 11: DecryptionS1JoinService and DecryptionS2AnnounceService are the meat of this PR, this is where the guardians calculate tally and spoiled ballot decryption shares and the guardian turns them into plaintext.

src/electionguard_gui/web/components/admin/admin-home-component.js
```diff
@@ -6,6 +6,21 @@ export default {
...
     ElectionsList,
   },
+  data() {
+    return {
+      loading: true,
+      keyCeremonies: [],
+    };
+  },
+  async mounted() {
+    const result = await eel.get_key_ceremonies()();
```
lprichar, line 16: In refactoring the guardian home page to pull all data so the watching operation could happen there, the admin home now also needs to pull data and push it down to the key ceremony component.  A nice side-effect is the admin never need to watch for updated key ceremonies, so now it doesn't.

src/electionguard_gui/components/guardian_home_component.py
```diff
@@ -0,0 +1,67 @@
...
+        js_key_ceremonies = [
+            key_ceremony.to_id_name_dict() for key_ceremony in key_ceremonies
+        ]
+        return eel_success(js_key_ceremonies)
+
+    def watch_db_collections(self) -> None:
+        self._log.debug("Watching database")
+        db = self._db_service.get_db()
+        self._db_watcher_service.watch_database(db, None, notify_ui_db_changed)
+        self._log.debug("exited watching key_ceremonies")
```
SteveMaier-IRT, line 55: not just watching key ceremonies here.  Could fix the text.

src/electionguard_gui/services/decryption_stages/decryption_s1_join_service.py
```diff
@@ -0,0 +1,45 @@
...
+        context = election.get_context()
+
+        guardian = self._guardian_service.load_guardian_from_decryption(
+            current_user_id, decryption
+        )
+        ballots = self._ballot_upload_service.get_ballots(db, election.id)
+        spoiled_ballots = [
+            ballot for ballot in ballots if ballot.state == BallotBoxState.SPOILED
+        ]
+        ciphertext_tally = get_tally(manifest, context, ballots)
```
SteveMaier-IRT, line 28: Just checking that the get_tally call can handle the ballots that are cast AND spoiled.

SteveMaier-IRT (approved):
looks good

src/electionguard_gui/services/decryption_stages/decryption_s1_join_service.py
lprichar: Thanks for double checking.  I had to think about it, but this is correct as is.  The `ballots` variable will contain all ballots both cast AND spoiled.

src/electionguard_gui/components/guardian_home_component.py
lprichar: good catch, I'm going to fix in a downstream PR if that's ok

PR #725, opened 2022-07-28
#697 Export Election Record

### LINKED ISSUE #697: ✨ Publish Election Record in GUI

### Suggestion

As an admin I want to publish an election record for an election so that it can be sent to a verifier

### Possible Implementation

_No response_

### Anything else?

_No response_

lprichar:
### Issue

Fixes #697 

### Description
Adds the ability to export an election record to the GUI after a decryption has been performed.

### Testing
1. Create an election
2. Perform a decryption

Once the decryption has been completed the following button will now show up on the decryption page:

image

3. Click the button, you should end up on the new export election record page:

image

4. Click the export button

Expected: you should get an encryption package at the expected location and the screen should look like this:

image


src/electionguard_gui/services/export_service.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+
+def get_drives() -> list[str]:
+    dl = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+    drives = [f"{d}:\\" for d in dl if os.path.exists(f"{d}:")]
```
SteveMaier-IRT, line 6: Do we want to list all drives or just the removable ones?

SteveMaier-IRT: considering that the devices only have 1 fixed drive and then thumb drives, this should be ok for now.  Thought for the future

lprichar: agreed, that will be a nice enhancement for the future

PR #732, opened 2022-08-01
Reuse guardians rather than recreating them during election creation

### LINKED ISSUE #731: 🐞 Mitre Verifier fails to validate public key in GUI generated election records

### Current Behavior

When you run an end-to-end test on the GUI results the [Mitre verifier](https://mitre.github.io/ElectionGuardVerifier.jl/) fails to validate the public key.  The problem is that the GUI is regenerating the guardians during election creation rather than reusing the values from the key ceremony.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

lprichar:
### Issue

Fixes #731 

### Description
This reuses the guardians data from the key ceremony rather than creating new guardians during election creation.

### Testing
Run an end-to-end test and validate it with the Mitre verifier.  There should be no problem with the public key.

### DIFF
--- a/src/electionguard_gui/components/create_election_component.py
+++ b/src/electionguard_gui/components/create_election_component.py
@@ -16,6 +16,7 @@ from electionguard_gui.services import (
     KeyCeremonyService,
     GuiSetupInputRetrievalStep,
     ElectionService,
+    GuardianService,
 )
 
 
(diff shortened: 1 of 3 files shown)


SteveMaier-IRT (approved):
clean fix for the issue.  looks good.

Issue #735, opened 2022-08-08
✨ Add Testing to GUI


lprichar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Suggestion

In order to make a deadline testing was temporarily suspended for the GUI project.  We must unit testing back in.  This may require some architectural changes to the GUI to retrofit, but as Michael Feathers says testable code is quality code, so the end result will be a more maintainable, malleable, and flexible codebase.

PR #736, opened 2022-08-08
GUI Testing

lprichar:
### Issue

Addresses #735 

### Description
Adds in unit tests for date time parsing, decryption status computation, and "can join" functionality for guardians during decryption.  This introduces the `pytest-mock` library for mocking objects dependent objects and isolating the area under test. 

### Testing
Run the tests, they should pass.


src/electionguard_gui/eel_utils.py
```diff
@@ -14,7 +15,9 @@ def utc_to_str(utc_dt: datetime) -> str:
     if not utc_dt:
         return ""
     local = convert_utc_to_local(utc_dt)
-    return local.strftime("%b %#d, %Y %#I:%M %p")
+    if os.name == "nt":
```
lprichar, line 18: this also fixes a bug where date time formatting works different on a Mac than on PC

tests/base_test_case.py
```diff
@@ -1,13 +1,21 @@
...
-
+from pytest import fixture
+from pytest_mock import MockerFixture
 from electionguard.constants import PrimeOption
 
 
 class BaseTestCase(TestCase):
     """Base Test Case for overriding environment variables."""
 
+    mocker: MockerFixture
```
lprichar, line 12: Expose a mocker object to all tests since it's cumbersome to do and requires a pylint disable

tests/unit/electionguard_gui/test_decryption_dto.py
```diff
@@ -0,0 +1,100 @@
...
+            }
+        )
+
+        # ACT
+        status = decryption_dto.get_status()
+
+        # ASSERT
+        self.assertEqual(status, "decryption complete")
+
+    @patch("electionguard_gui.services.authorization_service.AuthorizationService")
```
lprichar, line 55: This demonstrates how to instantiate a mock object that can be passed to another method as a parameter

tests/unit/electionguard_gui/test_election_dto.py
```diff
@@ -0,0 +1,28 @@
...
+from tests.base_test_case import BaseTestCase
+
+
+class TestElectionDto(BaseTestCase):
+    """Test the ElectionDto class"""
+
+    def test_get_status_with_no_guardians(self) -> None:
+        # ARRANGE
+        self.mocker.patch(
+            "electionguard_gui.models.election_dto.utc_to_str",
```
lprichar, line 13: This demonstrates how to mock a function that's imported in the method under test.  This string is non-intuitive and took me a while to figure out.  It's the method under test then the name of the function to mock.

tests/unit/electionguard_gui/test_election_dto.py
```diff
@@ -0,0 +1,28 @@
...
+                "created_at": datetime(2020, 2, 3, 7, 10, 10, 0, tzinfo=timezone.utc),
+            }
+        )
+
+        # ACT
+        result = election_dto.to_dict()
+
+        # ASSERT
+        self.assertEqual("ABC", result["id"])
+        self.assertEqual(result["created_at"], "Feb 3, 2022 2:10 PM")
```
SteveMaier-IRT, line 28: for the above 2 statements, one has the first value first and the second has the fixed value second.  If the first value needs to be first then the second assert needs to be fixed and the ones in the test_decryption_dto also need to be switched.

SteveMaier-IRT (approved):
looks good.

PR #744, opened 2022-08-12
#701 Dockerize GUI

### LINKED ISSUE #701: ✨ Dockerize GUI

### Suggestion

The GUI needs a dockerfile so it can be more easily deployed

### Possible Implementation

_No response_

### Anything else?

_No response_

lprichar:
### Issue

Fixes #701 

### Description

Caveats: First of all, this PR is purely additive from a functional perspective.  In other words, anyone running EGUI from source should see no differences when they create key ceremonies, elections, or run decryptions.  However, this does change the location of the guardian private keys, so updating to this code will break old elections.

What this adds is that now instead of running lots of individual commands to start the database and set environment variables, and start multiple gui sessions you can just run one command: `make start-egui` which will:

1. Build an image for an EGUI docker container based on Ubuntu
2. Spin up the two Mongo database containers
3. Spin up 3 EGUI containers: an admin and two guardians
4. Link the EGUI containers to the database for networking
5. Mount two host drives to the EGUI containers
  a. For persistent data storage, in the host at `egui_mnt/data`, in the container at `/egui_mnt/data`
  b. For data export, in the host at `egui_mnt/export/`, in the container at `/egui_mnt/export`
6. Expose the EGUI functionality to the host as follows:
  - http://localhost:12801/main.html for admin
  - http://localhost:12802/main.html for guardian1
  - http://localhost:12803/main.html for guardian2

As part of this PR the following EGUI startup options have been parameterized:
- port
- mode (uses `None` when run inside a container so it won't start a browser and Chrome won't need to be installed)
- host (uses `0.0.0.0` for the host to be able to reach eel inside a container)

This additionally reconfigures the database persistent storage to use a location that is relative to the project, and that is additionally the same location whether you use Docker or run EGUI locally so that you can switch seamlessly between the two modes of operation and your data will still work.

### Testing
This requires a full end-to-end test in both docker world and in from source world.  I have done both including validating with Julia, but would especially welcome validation that the docker containers work on someone else's machine.


.gitignore
```diff
@@ -154,3 +154,5 @@ sample-data.zip*
 
 results/
 gui_private_keys/
+database/
+egui_mnt/
```
lprichar, line 158: will fix no newline at end of file

src/electionguard_gui/.dockerignore
```diff
@@ -0,0 +1,2 @@
+docker-compose*.yml
+Dockerfile
```
lprichar, line 2: will fix

src/electionguard_gui/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+ENV PYENV_ROOT="/root/.pyenv"
+ENV PATH="$PYENV_ROOT/bin:$PYENV_ROOT/shims:${PATH}"
+RUN echo $PATH
+
+##################################################################################
+# Install python
+##################################################################################
+
+RUN pyenv install 3.9.9 && \
+    pyenv global 3.9.9
```
lprichar, line 37: Upgrading python will be easy, just adjust these two lines

src/electionguard_gui/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+
+COPY ./src ./src
+RUN rm src/electionguard_gui/__init__.py
+
+##################################################################################
+# Run EGUI
+##################################################################################
+
+# the final poetry install runs fast, it activates the virtualenv and initializes the modules
+RUN poetry install
```
lprichar, line 70: I tried to be careful to allow the image to cache as much as possible, but it still takes 10 seconds to build an image if anything changes.  There's room for improvement here, but it's certainly good enough.

SteveMaier-IRT (approved):
looks good.  the changes make the system much more configurable overall.

SteveMaier-IRT:
Please change the main.html to index.html so that we do not need to add it with the browser.

SteveMaier-IRT:
When you close a browser tab or change addresses in a tab that is pointing to an admin or guardian instance, the docker image exits.  This might be fixed by adding a restart flag to the container.

lprichar:
Good catches @SteveMaier-IRT, thanks, they're fixed now

Issue #739, opened 2022-08-09
🐞 MissingValueError on Guardian Decrypt


lprichar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Occasionally an error will occur when guardians are decrypting in the GUI.  Here's the relevant logs and stack trace:

```
2022-08-03 15:40:32.342135 DEBUG Starting main app
2022-08-03 15:40:32.342135 DEBUG Verifying database connection
2022-08-03 15:40:32.866290 DEBUG Starting eel
2022-08-03 15:40:35.405347 DEBUG Watching database
2022-08-03 15:40:36.197479 DEBUG Stopping watch database
2022-08-03 15:40:36.871910 DEBUG exited watching database
2022-08-03 15:40:42.730435 DEBUG Watching database
2022-08-03 16:22:33.088349 DEBUG new delta found for decryptions 62ead908b514c0521b66bd33
2022-08-03 16:22:38.743689 DEBUG Stopping watch database
2022-08-03 16:22:38.810225 DEBUG watching decryption '62ead908b514c0521b66bd33'
2022-08-03 16:22:46.705829 INFO  S1: 2 decrypting  62ead908b514c0521b66bd33
2022-08-03 16:22:46.803114 DEBUG loading guardian from C:\Users\eg202\documents\code\electionguard-python\gui_private_keys\62ea94cab65249d79ddd6ad0\guardian_2.json
2022-08-03 16:22:49.369442 ERROR missing value for field "contests.ballot_selections.proof.proof_one_data"
Traceback (most recent call last):
  File "C:\Users\eg202\documents\code\electionguard-python\.venv\lib\site-packages\dacite\core.py", line 59, in from_dict
    field_data = data[field.name]
KeyError: 'proof_one_data'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\eg202\documents\code\electionguard-python\.venv\lib\site-packages\dacite\core.py", line 71, in from_dict
    value = get_default_value_for_field(field)
  File "C:\Users\eg202\documents\code\electionguard-python\.venv\lib\site-packages\dacite\dataclasses.py", line 21, in get_default_value_for_field
    raise DefaultValueNotFoundError()
dacite.dataclasses.DefaultValueNotFoundError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #749, opened 2022-08-16
#739 Logging

lprichar:
### Issue

Supports #739

### Description
Couldn't repro the 739 bug, so this adds several features that will help us if it occurs again:

* If an error occurs during ballot decryption the system will now log it as an error and continue processing
* This updates GUI logging to combine it with the regular EG logging
* This updates GUI logging to send logs to the filesystem

### Testing
Run the GUI, you should see file system logs at `egui_mnt\data\logs`

### DIFF
--- a/src/electionguard/logs.py
+++ b/src/electionguard/logs.py
@@ -16,12 +16,17 @@ class ElectionGuardLog(Singleton):
     """
 
     __logger: logging.Logger
+    __stream_handler: logging.StreamHandler
 
     def __init__(self) -> None:
         super().__init__()
 
         self.__logger = logging.getLogger("electionguard")
-        self.__logger.addHandler(get_stream_handler())
+        # Pythong's logger will use the most restrictive of the logger level and the handler level,
+        #   so set the logger to the lowest level the handler ever might log at
+        self.__logger.setLevel(logging.DEBUG)
+        self.__stream_handler = get_stream_handler(logging.INFO)
+        self.__logger.addHandler(self.__stream_handler)
 
     @staticmethod
     def __get_call_info() -> Tuple[str, str, int]:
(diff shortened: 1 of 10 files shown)


src/electionguard/logs.py
```diff
@@ -44,6 +49,13 @@ def __formatted_message(self, message: str) -> str:
         message = f"{os.path.basename(filename)}.{funcname}:#L{line}: {message}"
         return message
 
+    def set_stream_log_level(self, Level: int) -> None:
```
lprichar, line 52: Previously the stream log level was hard-coded to INFO.  It is now settable since GUI will log at DEBUG.

src/electionguard/logs.py
```diff
@@ -16,12 +16,17 @@ class ElectionGuardLog(Singleton):
...
+    __stream_handler: logging.StreamHandler
 
     def __init__(self) -> None:
         super().__init__()
 
         self.__logger = logging.getLogger("electionguard")
-        self.__logger.addHandler(get_stream_handler())
+        # Pythong's logger will use the most restrictive of the logger level and the handler level,
+        #   so set the logger to the lowest level the handler ever might log at
+        self.__logger.setLevel(logging.DEBUG)
```
lprichar, line 27: EG used to only log ERROR and above even though it was set at INFO, because logger has a level at both the logger and handler levels, it uses the most restrictive of those two, and if the logger isn't set it will use ERROR.

src/electionguard/logs.py
```diff
@@ -16,12 +16,17 @@ class ElectionGuardLog(Singleton):
...
+        self.__stream_handler = get_stream_handler(logging.INFO)
```
lprichar, line 28: Because of the bug above EG was asking for INFO but only getting ERROR.  My bug fix above may mean that EG is now too verbose.  May need to adjust this ^ if so.

SteveMaier-IRT (approved):
lgtm

PR #750, opened 2022-08-17
#743 Fix Guardian ids in Lagrange Coefficients

### LINKED ISSUE #743: 🐞 Guardians must be named with numbers

### Current Behavior

Otherwise the election record will fail to validate

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

lprichar:
### Issue

Fixes #743 

### Description
Guardians previously needed to be named with numbers else the election verification would fail when validating the Lagrange Coefficients stating that the guardian_id needed to be numeric.  This fixes it by having the admin look up the guardian's sequence order and announcing guardians with that value instead of the guardian's alpha-numeric identifier.

### Testing
Run an end-to-end election with guardians with alpha-numeric strings and validate it in Julia.  Expected: it should validate.

### DIFF
--- a/src/electionguard_gui/models/election_dto.py
+++ b/src/electionguard_gui/models/election_dto.py
@@ -105,6 +105,12 @@ class ElectionDto:
             raise Exception("No guardian records found")
         return from_list_raw(GuardianRecord, self.guardian_records)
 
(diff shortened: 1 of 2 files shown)


SteveMaier-IRT (approved):
lgtm

PR #771, opened 2022-08-24
#768 - System now successfully resumes from sleep on guardian page

### LINKED ISSUE #768: 🐞 GUI Dies After Several Hours of Running

### Current Behavior

If you let the app run for an extended amount of time, eventually it dies

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

lprichar:
### Issue

Fixes #768 

### Description
The underlying issue was that Eel in Python would die when the system was resuming from sleep.  The web-based UI then looked like it was still working when in fact the server had crashed.  That gave the appearance of key ceremonies and decryptions not appearing on guardian devices when they should have, when in fact the issue was much larger.

The server-side solution is to disallow eel from ever exiting of its own accord.  This allows eel to resume from system sleep, although it has the consequence that the user must now explicitly hit Ctrl+C to end the app.

The client-side solution is to watch for system resume events and when they occur refresh the page.  This solution also adds a refresh button for convenience, although the prior solutions should warrant it unnecessary.

### Testing
1. Run the app as a guardian
2. Sleep/hibernate your system
3. Have an admin add a key ceremony

Expected the guardian should see the new key ceremony on their home page

### DIFF
--- a/src/electionguard_gui/components/guardian_home_component.py
+++ b/src/electionguard_gui/components/guardian_home_component.py
@@ -49,10 +49,18 @@ class GuardianHomeComponent(ComponentBase):
         return eel_success(js_key_ceremonies)
 
     def watch_db_collections(self) -> None:
-        self._log.debug("Watching database")
-        db = self._db_service.get_db()
-        self._db_watcher_service.watch_database(db, None, notify_ui_db_changed)
-        self._log.debug("exited watching database")
+        try:
+            self._log.debug("Watching database")
+            db = self._db_service.get_db()
+            self._db_watcher_service.watch_database(db, None, notify_ui_db_changed)
+            self._log.debug("exited watching database")
+        except KeyboardInterrupt:
+            self._log.debug("Keyboard interrupt, exiting watch database")
+            self._db_watcher_service.stop_watching()
+        except Exception as e:  # pylint: disable=broad-except
+            self.handle_error(e)
+            self._db_watcher_service.stop_watching()
+            # no need to raise exception or return anything, we're in fire-and-forget mode here
 
     def stop_watching_db_collections(self) -> None:
         self._log.debug("Stopping watch database")
--- a/src/electionguard_gui/main_app.py
+++ b/src/electionguard_gui/main_app.py
@@ -116,4 +116,4 @@ class MainApp:
             raise e
 
     def on_close(self, _page: str, _open_sockets: list) -> None:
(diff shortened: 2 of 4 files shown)


src/electionguard_gui/main_app.py
```diff
@@ -101,8 +101,19 @@ def start(self) -> None:
...
             host = self.config_service.get_host()
             self.log_service.debug(f"Starting eel port={port} mode={mode} host={host}")
-            eel.start("index.html", size=(1024, 768), port=port, mode=mode, host=host)
+            eel.start(
+                "index.html",
+                size=(1024, 768),
+                port=port,
+                mode=mode,
+                host=host,
+                close_callback=self.on_close,
```
lprichar, line 110: simply overriding the `close_callback` method doesn't seem like it should stop the eel from exiting on it's own, but whether it's a bug or not that's how it actually works

src/electionguard_gui/components/guardian_home_component.py
```diff
@@ -49,10 +49,18 @@ def get_key_ceremonies(self) -> dict[str, Any]:
...
-        self._log.debug("Watching database")
-        db = self._db_service.get_db()
-        self._db_watcher_service.watch_database(db, None, notify_ui_db_changed)
-        self._log.debug("exited watching database")
+        try:
+            self._log.debug("Watching database")
+            db = self._db_service.get_db()
+            self._db_watcher_service.watch_database(db, None, notify_ui_db_changed)
+            self._log.debug("exited watching database")
+        except KeyboardInterrupt:
```
SteveMaier-IRT, line 57: Do we need anything here for touchscreen/mouse interrupts as well?

SteveMaier-IRT (changes requested):
When the admin times out because the tablet was put to sleep using the power button, the UI will not show the existing elections.  The console will tell the user to hit ctrl+C to close the egui app, but this does not close the gui at all.

The refresh button on the guardian home page worked well and was able to reconnect from a lost socket.

The admin pages are abit more complex and a solution will be needed from multiple pages, not just the home page like the guardians.

lprichar:
Great points.  Added issue to track remaining work:

https://github.com/microsoft/electionguard-python/issues/773

I will fix the Ctrl+C comment in this PR shortly.

SteveMaier-IRT (approved):
lgtm for the guardian

PR #774, opened 2022-08-25
#766 Ballot Enhancements

### LINKED ISSUE #766: 🐞 Upload Ballots Enhancements

### Current Behavior

During uploading of ballots, the ballots are not checked against the manifest or anything else for that election.  This has led during testing to ballots from the wrong election being uploaded.  This then makes the tally process not able to complete.

### Expected Behavior

Ballots should be verified on upload that they are part of the election.

If the encryption is also checked at this time, then the tally on the admin side would not need to check the encryption on each tally run.

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

lprichar:
### Issue
Fixes #766 

### Description
Primarily this adds additional validation during ballot upload that the ballots match the manifest.  It also:

- Fixes an issue where error messages weren't displayed correctly on the decryption page
- Improves the UI of the election page
  - Moves buttons on the menu to appropriate sections where possible
  - Displays a nice message if there aren't any ballots and tallies
  - Moves the title down and separates the sections more clearly
  - Updates button colors from green to gray
  - Move title lower on page and adjusts spacing
  - Add tally date to tallies list
  - Add upload date to ballots list
- Improves the UI of the decryption page
  - Adjusts UI to match election page e.g. buttons to grey
  - Renames "Decryption Results" to "Tally Results"
  - Moves spoiled ballots into a table
  - Show a nice message if no spoiled ballots exist

### Testing

Try uploading a ballot that belongs to a different manifest, it should now show like this:

image

A bran new election with no ballots uploaded looks like this:

image

An election with a few ballot uploads and a tally now looks like this:

image

### DIFF
--- a/src/electionguard_gui/web/components/admin/view-decryption-admin-component.js
+++ b/src/electionguard_gui/web/components/admin/view-decryption-admin-component.js
@@ -72,13 +72,10 @@ export default {
       <p class="alert alert-danger" role="alert">An error occurred. Check the logs and try again.</p>
     </div>
     <div v-if="decryption">
-      <div class="row mb-4">
-        <div class="col-11">
-          <h1>{{decryption.decryption_name}}</h1>
-        </div>
-        <div class="col-1 text-end" v-if="decryption.completed_at_str">
+      <div class="text-end">
+        <div v-if="decryption.completed_at_str">
           <div class="dropdown">
-            <button class="btn btn-primary dropdown-toggle" type="button" data-bs-toggle="dropdown" aria-expanded="false">
+            <button class="btn btn-sm btn-secondary dropdown-toggle" type="button" data-bs-toggle="dropdown" aria-expanded="false">
               <i class="bi-gear-fill me-1"></i>
             </button>
             <ul class="dropdown-menu">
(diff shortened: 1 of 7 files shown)


src/electionguard_gui/web/components/admin/view-decryption-admin-component.js
```diff
@@ -29,26 +29,37 @@ export default {
...
-      if (result.success) {
-        this.decryption = result.result;
+      try {
+        const result = await eel.get_decryption(
+          this.decryptionId,
+          is_refresh
+        )();
+        this.error = !result.success;
+        if (result.success) {
+          this.decryption = result.result;
```
SteveMaier-IRT, line 50: on a failure, we probably want to set the this.decryption back to an initial state so that we are not using a different one.

lprichar: An excellent point in general.  In this particular instance, however, setting the decryption to null would make most of the UI disappear because of the v-if on line 74.  I'm not sure we necessarily want that.  It would require the user to either hit the back button (alt-left) or Ctrl+R to reload the app, or to add some UI to handle that scenario.

SteveMaier-IRT (approved):
lgtm

PR #778, opened 2022-08-26
#762 Version Info

### LINKED ISSUE #762: ✨ Print Version Info in GUI

### Suggestion

Currently the GUI doesn't give any indication of what version of code it's running.  Ideally it would show something like a date of last commit and/or commit message if it's running in a git enabled folder, else date of last modification if it's in docker.

### Possible Implementation

_No response_

### Anything else?

_No response_

lprichar:
### Issue
Fixes #762

### Description
Shows the version info when you run via `git egui` (not via docker, although we could add that with very little effort)

### Testing
1. Run `git egui`

Expected, a git sha hash like this:

image

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -373,28 +373,6 @@ python-dateutil = ">=2.8.1"
 dev = ["twine", "markdown", "flake8", "wheel"]
 
 [[package]]
-name = "gitdb"
-version = "4.0.9"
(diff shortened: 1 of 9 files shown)


SteveMaier-IRT (approved):
lgtm.  Love the use of git commit hash.

PR #786, opened 2022-09-16
Electionguard-python

CodingForFun-coder:
Co-authored-by: JPankey-Thompson <JKnight777@users.noreply.github.com>
We created a button, "generate pdf", when pressed will print out the summary of the data that had been collected during the voting process.

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,2 @@
 EG_DB_DIR=../../database
+EG_DB_PASSWORD= #Password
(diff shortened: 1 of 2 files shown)


.env
```diff
@@ -1 +1,2 @@
-EG_DB_DIR=../../database
\ No newline at end of file
+EG_DB_DIR=../../database
+EG_DB_PASSWORD= #Password
```
lprichar, line 2: Please remove the password from the PR, it isn't related to pdf generation

lprichar (changes requested):
I _love_ the simplicity of this solution, great job!  Please just remove the password from the .env file.  That's a problem we need to solve, but not here.

lprichar (approved):
lgtm!

